PR #11, opened 2017-07-24
Apply checkstyle rules on whole repository

sutaakar:


pom.xml
```diff
@@ -24,6 +24,8 @@
 
   <properties>
     <findbugs.failOnViolation>true</findbugs.failOnViolation>
+    <checkstyle.failOnViolation>true</checkstyle.failOnViolation>
+    <checkstyle.logViolationsToConsole>true</checkstyle.logViolationsToConsole>
```
jhrcek: FYI: As pointed out by Bojan today, this property has no effect: https://github.com/kiegroup/jbpm-wb/pull/807#discussion_r128999897

sutaakar: @jhrcek Actually this property has an effect, see https://github.com/kiegroup/jbpm-wb/pull/807#issuecomment-317477600

PR #16, opened 2017-07-25
Test stabilizations

sutaakar:


framework-cloud/framework-cloud-common/src/main/java/org/kie/cloud/common/provider/KieServerMgmtClientProvider.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public KieServerMgmtControllerClient getKieServerMgmtControllerClient() {
+        return kieServerMgmtControllerClient;
+    }
+
+    /**
+     * Wait until any server template is created in controller.
+     */
+    public void waitForServerTemplateCreation() {
+        long timeoutTime = Calendar.getInstance().getTimeInMillis() + 30000L;
```
jiripetrlik: Would it be possible to use new Date time API in Java 8?

sutaakar: Good catch, updated.

framework-cloud/framework-cloud-common/src/main/java/org/kie/cloud/common/provider/KieServerMgmtClientProvider.java
```diff
@@ -0,0 +1,60 @@
...
+package org.kie.cloud.common.provider;
+
+import java.time.Instant;
+import java.util.Collection;
+
+import org.kie.cloud.api.deployment.WorkbenchDeployment;
+import org.kie.server.controller.api.model.spec.ServerTemplate;
+import org.kie.server.integrationtests.controller.client.KieServerMgmtControllerClient;
+
+public class KieServerMgmtClientProvider {
```
jhrcek: Why call it KieServerMgmtClientProvider if it provides KieServerMgmtControllerClient?
It'd prefer KieServerControllerClientProvider

framework-cloud/framework-cloud-common/src/main/java/org/kie/cloud/common/provider/KieServerMgmtClientProvider.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            try {
+                Thread.sleep(200L);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Interrupted while waiting for server template creation.", e);
+            }
+        }
+        throw new RuntimeException("Timeout while waiting for server template creation.");
```
jhrcek: I usually include timeout time in the exception message to make it clear without looking at the sources how long was the time (how slow it was).

Like "Timeout while waiting for 30 seconds for server template creation" or such

framework-cloud/framework-cloud-common/src/main/java/org/kie/cloud/common/provider/KieServerMgmtClientProvider.java
jakubschwan: Better, but longer description of this class can be KieServerMgmtControllerClientProvider

sutaakar: I think that KieServerControllerClientProvider should be enough, adjusted.

framework-cloud/framework-cloud-common/src/main/java/org/kie/cloud/common/provider/KieServerMgmtClientProvider.java
sutaakar: Done.

jiripetrlik:
Looks good for me now.

PR #18, opened 2017-07-26
Fix and little refactoring of cloud deployment plugin

jiripetrlik:
- All cloud-api implementations will be added to cloud-deployment-plugin dependencies
- Implementation can be chosen using "implementation" parameter

### DIFF
--- a/framework-cloud/cloud-deployment-plugin/pom.xml
+++ b/framework-cloud/cloud-deployment-plugin/pom.xml
@@ -38,5 +38,10 @@
       <artifactId>maven-plugin-annotations</artifactId>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
+      <groupId>org.kie.cloud</groupId>
+      <artifactId>framework-openshift</artifactId>
+    </dependency>
   </dependencies>
 </project>
--- a/framework-cloud/cloud-deployment-plugin/src/main/java/org/kie/cloud/plugin/Undeploy.java
+++ b/framework-cloud/cloud-deployment-plugin/src/main/java/org/kie/cloud/plugin/Undeploy.java
@@ -27,7 +27,9 @@ import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.kie.cloud.api.DeploymentScenarioBuilderFactory;
+import org.kie.cloud.api.DeploymentScenarioBuilderFactoryLoader;
 
+import static org.kie.cloud.plugin.Constants.CLOUD_API_IMPLEMENTATION_PROPERTY;
 import static org.kie.cloud.plugin.Constants.NAMESPACE_PROPERTY;
 import static org.kie.cloud.plugin.Constants.PROPERTY_FILE_PATH;
 
@@ -38,11 +40,12 @@ public class Undeploy extends AbstractMojo {
     private MavenProject mavenProject;
 
     @Override public void execute() throws MojoExecutionException, MojoFailureException {
-        DeploymentScenarioBuilderFactory deploymentScenarioBuilderFactory = DeploymentScenarioBuilderFactory.getInstance();
-        deploymentScenarioBuilderFactory.deleteNamespace(getNamespace());
+        DeploymentScenarioBuilderFactory deploymentScenarioBuilderFactory = DeploymentScenarioBuilderFactoryLoader
+                .getInstance(loadCloudProperties().getProperty(CLOUD_API_IMPLEMENTATION_PROPERTY));
+        deploymentScenarioBuilderFactory.deleteNamespace(loadCloudProperties().getProperty(NAMESPACE_PROPERTY));
     }
 
-    private String getNamespace() {
+    private Properties loadCloudProperties() {
         Properties properties = new Properties();
         try (InputStream inputStream = new FileInputStream(mavenProject.getModel().getBuild().getOutputDirectory() + PROPERTY_FILE_PATH)) {
             properties.load(inputStream);
@@ -50,6 +53,6 @@ public class Undeploy extends AbstractMojo {
             throw new RuntimeException("Error loading Openshift properties", e);
         }
 
-        return properties.getProperty(NAMESPACE_PROPERTY);
+        return properties;
     }
 }
--- a/framework-cloud/cloud-deployment-plugin/src/main/java/org/kie/cloud/plugin/Undeploy.java
+++ b/framework-cloud/cloud-deployment-plugin/src/main/java/org/kie/cloud/plugin/Undeploy.java
@@ -15,6 +15,7 @@
 
 package org.kie.cloud.plugin;
 
+import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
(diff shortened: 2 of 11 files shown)


framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/DeploymentScenarioBuilderFactoryLoader.java
```diff
@@ -0,0 +1,60 @@
...
+public class DeploymentScenarioBuilderFactoryLoader {
+    public static DeploymentScenarioBuilderFactory getInstance() {
+        List<DeploymentScenarioBuilderFactory> deploymentScenarioBuilderFactories = getFactories();
+        if (deploymentScenarioBuilderFactories.size() == 1) {
+            return deploymentScenarioBuilderFactories.get(0);
+        } else if (deploymentScenarioBuilderFactories.size() == 0) {
+            throw new RuntimeException("No cloud API implementation was found");
+        } else {
+            StringBuilder cloudImplementations = new StringBuilder();
+            deploymentScenarioBuilderFactories.stream().forEach(x -> cloudImplementations.append(x.getCloudEnvironmentName() + ", "));
```
sutaakar: Could be adjusted to deploymentScenarioBuilderFactories.stream().map(x -> x.getCloudEnvironmentName()).collect(Collectors.joining(", "));

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/DeploymentScenarioBuilderFactoryLoader.java
```diff
@@ -0,0 +1,60 @@
...
+
+            throw new RuntimeException("Multiple cloud API implementations detected - please select one from: " + cloudImplementations.toString());
+        }
+    }
+
+    public static DeploymentScenarioBuilderFactory getInstance(String environmentName) {
+        List<DeploymentScenarioBuilderFactory> deploymentScenarioBuilderFactories = getFactories();
+        Map<String, DeploymentScenarioBuilderFactory> factoryMap = new HashMap<>();
```
sutaakar: Could be adjusted to "return deploymentScenarioBuilderFactories.stream().filter(x -> x.getCloudEnvironmentName().equals(environmentName)).findAny().orElseThrow(() -> new RuntimeException("No implementantion of cloud API with name " + environmentName + " was found"));"

framework-cloud/cloud-deployment-plugin/pom.xml
```diff
@@ -38,5 +38,10 @@
       <artifactId>maven-plugin-annotations</artifactId>
       <scope>provided</scope>
     </dependency>
+
+    <dependency>
```
sutaakar: Should we directly add OpenShift dependency to the plugin?
Or we should rather leave this to the plugin users? So every user will add cloud dependency to the plugin based on his choice?

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/DeploymentScenarioBuilderFactoryLoader.java
jiripetrlik: Thanks! Good tip! It was changed.

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/DeploymentScenarioBuilderFactoryLoader.java
jiripetrlik: Thank you for nice stream. But I'm not sure about this. I don't like huge streams which replace whole methods.

framework-cloud/cloud-deployment-plugin/pom.xml
jiripetrlik: It is problem in this case because this plugin will be often used externally. So, I think it would be better to add all dependencies to its pom and allow user to select cloud environment using parameter in Maven pom.

framework-cloud/cloud-deployment-plugin/src/main/java/org/kie/cloud/plugin/Undeploy.java
```diff
@@ -38,18 +40,19 @@
...
     @Override public void execute() throws MojoExecutionException, MojoFailureException {
-        DeploymentScenarioBuilderFactory deploymentScenarioBuilderFactory = DeploymentScenarioBuilderFactory.getInstance();
-        deploymentScenarioBuilderFactory.deleteNamespace(getNamespace());
+        DeploymentScenarioBuilderFactory deploymentScenarioBuilderFactory = DeploymentScenarioBuilderFactoryLoader
+                .getInstance(loadCloudProperties().getProperty(CLOUD_API_IMPLEMENTATION_PROPERTY));
+        deploymentScenarioBuilderFactory.deleteNamespace(loadCloudProperties().getProperty(NAMESPACE_PROPERTY));
     }
 
-    private String getNamespace() {
+    private Properties loadCloudProperties() {
```
jhrcek: I don't like that instead of domain specific properties (like namespace and now the newly added cloud api implementation property) we're now returning the entire properties.

Wouldn't it be better to encapsulate the properties in some more domain specific class, to let the api user know what properties are retrievable? Otherwise api user will have to look at usages of this method and search what is being stored into it and then also look for constants representing valid keys etc.

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/DeploymentScenarioBuilderFactory.java
```diff
@@ -16,22 +16,13 @@
 package org.kie.cloud.api;
 
 import java.util.Iterator;
+import java.util.List;
 import java.util.ServiceLoader;
 
 import org.kie.cloud.api.scenario.builder.WorkbenchWithKieServerScenarioBuilder;
 
 public interface DeploymentScenarioBuilderFactory {
+    String getCloudEnvironmentName();
```
jhrcek: cloud environment name vs. CLOUD_API_IMPLEMENTATION constant - lets unify the naming, please

framework-cloud/cloud-deployment-plugin/src/main/java/org/kie/cloud/plugin/Undeploy.java
jiripetrlik: Changed to better name.

jiripetrlik: Separate class for plugin settings was created - "CloudDeploymentPluginConfiguration".

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/DeploymentScenarioBuilderFactory.java
jiripetrlik: Method was renamed.

jiripetrlik:
@jhrcek 
Thank you for your review. Can you please look at it again.

PR #24, opened 2017-08-02
[GUVNOR-3364] Workbench HTTPS tests

jiripetrlik:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/WorkbenchWithKieServerScenarioImpl.java
```diff
@@ -87,6 +88,13 @@ public WorkbenchWithKieServerScenarioImpl(OpenShiftController openShiftControlle
         } catch (MalformedURLException e) {
             throw new RuntimeException("Malformed URL for workbench", e);
         }
+        String secureRouteHostWorkbench = project.getService(workbenchDeployment.getSecureServiceName()).getRoute().getRouteHost();
+        String secureUrlWorkbench = "https://" + secureRouteHostWorkbench + "/kie-wb.jsp";
```
sutaakar: Does it work for REST tests? I am quite surprised that this "/kie-wb.jsp" works with REST.

jiripetrlik: Thank you. Good catch. It worked because "/kie-wb.jsp" was truncated in costructor "new URL()". However, it is fixed now.

PR #27, opened 2017-08-04
Separate common test functionality

sutaakar:
Adjustment to simplify tests - common functionality moved to abstract parent.

### DIFF
--- a/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/https/WorkbenchHttpsIntegrationTest.java
+++ b/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/https/WorkbenchHttpsIntegrationTest.java
@@ -39,46 +39,29 @@ import org.apache.http.impl.client.CloseableHttpClient;
 import org.apache.http.impl.client.HttpClients;
 import org.apache.http.ssl.SSLContextBuilder;
 import org.assertj.core.api.Assertions;
-import org.junit.After;
-import org.junit.Before;
 import org.junit.Test;
 import org.kie.cloud.api.DeploymentScenarioBuilderFactory;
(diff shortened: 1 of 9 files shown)


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/https/WorkbenchHttpsIntegrationTest.java
```diff
@@ -39,46 +39,29 @@
...
-        workbenchWithKieServerScenario.deploy();
-    }
-
-    @After
-    public void tearDown() {
-        InstanceLogUtil.writeDeploymentLogs(workbenchWithKieServerScenario.getWorkbenchDeployment());
-        InstanceLogUtil.writeDeploymentLogs(workbenchWithKieServerScenario.getKieServerDeployment());
-        workbenchWithKieServerScenario.undeploy();
+    @Override
+    protected WorkbenchWithKieServerScenario createDeploymentScenario(DeploymentScenarioBuilderFactory deploymentScenarioFactory) {
```
jiripetrlik: Can we somehow avoid of this abstract method? The information about which deployment scenario use is already present in parameter T.

sutaakar: We can provide default implementation in AbstractCloudIntegrationTest. In that case only the tests who needs specific configuration will override the method implementation.
But this will require changes in cloud API. I think we should implement that in separate PR.

jiripetrlik: Well, I agree. Otherwise it looks good for me.

PR #35, opened 2017-08-16
Bulk deployment initialization

sutaakar:
Initialize deployments in a bulk to have all informations available even in case of failure while scaling.

### DIFF
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/WorkbenchWithKieServerScenarioImpl.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/WorkbenchWithKieServerScenarioImpl.java
@@ -103,9 +103,6 @@ public class WorkbenchWithKieServerScenarioImpl implements WorkbenchWithKieServe
             throw new RuntimeException("Malformed secure URL for workbench", e);
         }
 
-        logger.info("Waiting for Workbench deployment to become ready.");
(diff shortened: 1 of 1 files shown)


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/WorkbenchWithKieServerScenarioImpl.java
```diff
@@ -129,18 +126,21 @@ public void deploy() {
...
         databaseDeployment.setOpenShiftController(openshiftController);
         databaseDeployment.setNamespace(projectName);
         databaseDeployment.setDatabaseName(DeploymentConstants.getDatabaseName());
         databaseDeployment.setApplicationName(OpenShiftConstants.getKieApplicationName());
 
+        logger.info("Waiting for Workbench deployment to become ready.");
+        workbenchDeployment.waitForScale();
+
+        logger.info("Waiting for Kie server deployment to become ready.");
+        kieServerDeployment.waitForScale();
```
jakubschwan: Database deployment should wait for scale before Kie server deployment. I think that this should be safer - Kie Server has database ready and can connect to it.

sutaakar: Hmm, it can be so, topic for another PR :)

PR #40, opened 2017-08-22
Derive database service name from available services

sutaakar:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/DatabaseDeploymentImpl.java
```diff
@@ -90,6 +94,17 @@ public String getNamespace() {
     }
 
     public String getServiceName() {
+        if (databaseName == null) {
+            // Database name not set, try to guess it from all available services
+            Pattern pattern = Pattern.compile(DATABASE_REGEXP);
```
jhrcek: I'd initialize the regex directly in the static field:
`private static final Pattern DATABASE_REGEXP = Pattern.compile(...)`
but leave as it is if you prefer this way.

sutaakar: Adjusted, please merge if everything is ok.

PR #41, opened 2017-08-24
Create separate scenario for Workbench with Kie server

sutaakar:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/constants/OpenShiftConstants.java
```diff
@@ -29,7 +29,9 @@
 
     public static final String KIE_APP_SECRET = "kie.app.secret";
     public static final String KIE_IMAGE_STREAMS = "kie.image.streams";
-    public static final String KIE_APP_TEMPLATE = "kie.app.template";
+    public static final String KIE_APP_TEMPLATE_WORKBENCH_KIE_SERVER_DATABASE = "kie.app.template.workbench.kie-server.database";
```
jhrcek: What type of thing do these 3 properties point to? Is that some url, some local file?

I'm kind of concerned that we're encoding a log of information in plain Strings here. Bunch of properties for which it's not clear what should I set as a value for them. Can we at least have this documented in javadocs?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/WorkbenchWithKieServerInDifferentNamespaceScenarioImpl.java
```diff
@@ -0,0 +1,186 @@
...
+
+        String routeHostWorkbench = workbenchProject.getService(workbenchDeployment.getServiceName()).getRoute().getRouteHost();
+        String urlWorkbench = "http://" + routeHostWorkbench;
+        try {
+            workbenchDeployment.setUrl(new URL(urlWorkbench));
+        } catch (MalformedURLException e) {
+            throw new RuntimeException("Malformed URL for workbench", e);
+        }
+
+        kieServerDeployment = new KieServerDeploymentImpl();
```
jhrcek: This method is kind of big. Could you please split it somehow (either two methods one for each project or three - one for each deployment?)

sutaakar:
Closing as this scenario won't be implemented. There are going to be just scenarios where all deployments are placed within one project and special scenario with every deployment in special project

PR #43, opened 2017-08-30
Change maven property for OpenShift template

sutaakar:


test-cloud/pom.xml
```diff
@@ -23,7 +23,10 @@
 
     <kie.image.streams/> <!-- Needs to be defined for proper test run. -->
     <kie.app.secret>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/secrets/bpmsuite-app-secret.json</kie.app.secret>
-    <kie.app.template>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-full-mysql-persistent.json</kie.app.template>
+    <kie.app.template.workbench.kie-server.database>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-full-mysql-persistent.json</kie.app.template.workbench.kie-server.database>
```
jiripetrlik: It would be good to move references on templates somewhere outside pom.xml. These properties may change with each build.

PR #44, opened 2017-08-30
Move template definitions to properties files

sutaakar:
To run tests using database (currently all tests) you need to add proper profile to mvn command.

### DIFF
--- a/test-cloud/pom.xml
+++ b/test-cloud/pom.xml
@@ -23,10 +23,11 @@
 
     <kie.image.streams/> <!-- Needs to be defined for proper test run. -->
     <kie.app.secret>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/secrets/bpmsuite-app-secret.json</kie.app.secret>
-    <kie.app.template.workbench.kie-server.database>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-full-mysql-persistent.json</kie.app.template.workbench.kie-server.database>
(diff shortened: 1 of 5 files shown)


test-cloud/pom.xml
```diff
@@ -23,10 +23,11 @@
...
-    <kie.app.template.workbench.kie-server.database>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-full-mysql-persistent.json</kie.app.template.workbench.kie-server.database>
-    <kie.app.template.kie-server>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-executionserver.json</kie.app.template.kie-server>
-    <kie.app.template.kie-server.database>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-executionserver-postgresql-persistent.json</kie.app.template.kie-server.database>
-    <kie.app.template.workbench-monitoring.smartrouter>https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-businesscentral-monitoring-ephemeral.json</kie.app.template.workbench-monitoring.smartrouter>
+    <!-- Templates are specified in specific properties files. Path is relative to submodules of this module. -->
+    <kie.app.templates.general>../src/main/resources/templates-general.properties</kie.app.templates.general>
```
jhrcek: Maybe it's ok like this, but there are also built-in maven properties like:
`${project.parent.build.resources}` which you could use to avoid hardcoding paths to property files.

To check, that the path makes sense in the current submodule just check it like this:

```
cd test-cloud
mvn help:evaluate
//now you can interactively enter things like ${project.parent.build.resources} to see what it evaluates to in current project
```

sutaakar: @jhrcek Good tip, adjusted.

PR #51, opened 2017-09-13
Enhancements related to OpenShift routes

sutaakar:
This comes handy when you need to know service route before route is created, for example if route URL needs to be in environment variables.

### DIFF
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerDeploymentImpl.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerDeploymentImpl.java
@@ -36,19 +36,17 @@ public class KieServerDeploymentImpl extends OpenShiftDeployment implements KieS
 
 
     @Override public URL getUrl() {
+        if (url == null) {
+            url = getHttpRouteUrl(serviceName);
+        }
         return url;
     }
 
-    public void setUrl(URL url) {
-        this.url = url;
-    }
-
     @Override public URL getSecureUrl() {
+        if (secureUrl == null) {
+            secureUrl = getHttpsRouteUrl(secureServiceName);
+        }
         return secureUrl;
-    }
-
-    public void setSecureUrl(URL secureUrl) {
-        this.secureUrl = secureUrl;
     }
 
     @Override public String getUsername() {
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerDeploymentImpl.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerDeploymentImpl.java
@@ -106,8 +106,4 @@ public class KieServerDeploymentImpl extends OpenShiftDeployment implements KieS
             RouterUtil.waitForRouter(getUrl());
         }
     }
-
-    @Override public boolean ready() {
-        throw new RuntimeException("Not implemented");
-    }
 }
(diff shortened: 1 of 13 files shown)


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
```diff
@@ -63,4 +66,56 @@ public void scale(int instances) {
...
+
+    protected URL getHttpsRouteUrl(String serviceName) {
+        return getRouteUrl("https", serviceName);
+    }
+
+    private URL getRouteUrl(String protocol, String serviceName) {
+        URL url;
+        Service service = openShiftController.getProject(namespace).getService(serviceName);
+
+        StringBuffer urlValue = new StringBuffer();
```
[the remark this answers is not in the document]
triceo: When we're talking about `StringBuffer` v. `StringBuilder`, you are right.

But lately, `javac` has been replacing `String` concatenation with `StringBuilder` on the fly - so the cleanest, cheapest way once again is to just use `+`. Also, in Java 9 I think, they're creating new bytecode for this, so the JVM will be able to pick the best implementation on the fly.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerDeploymentImpl.java
```diff
@@ -36,21 +36,19 @@
 
 
     @Override public URL getUrl() {
+        if (url == null) {
+            url = getHttpRouteUrl(serviceName);
```
jakubschwan: What if we set URL for Kie server later, that doesn't match to created URL using default subdomain?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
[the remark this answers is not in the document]
jiripetrlik: Does it make sense to discuss performance in this case? Concatenating strings is quite cheap operation. And this is even not a loop or something. I would prefer to use the most readable way which is to use simple "+".

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/resource/OpenShiftResourceConstants.java
```diff
@@ -30,7 +30,6 @@
     public static final String DEPLOYMENT_CONFIG_LABEL = "deploymentConfig";
 
     // Route constants
-    public static final String CENTRAL_CI_ROUTE_SUFFIX = ".project.openshiftdomain";
```
jiripetrlik: Good catch. This Central CI dependent constant should not be present in this repo.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerDeploymentImpl.java
sutaakar: You mean the case when we need to know the URL in advance and we know that it will be different from default subdomain?
It that case we could reintroduce setUrl method where we can specify the URL. I deleted it because I don't expect that we will cover such scenario and I want to keep the code as simple as possible.

jakubschwan: +1 for simple code.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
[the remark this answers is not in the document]
sutaakar: Adjusted to use '+'

PR #52, opened 2017-09-13
GUVNOR-3484: Test for managed Kie server with router and console

sutaakar:
Implements Architecture 3 from http://mswiderski.blogspot.cz/2017/08/cloud-runtime-architectures-for-jbpm.html

### DIFF
--- a/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/emptymanagedkieserver/EmptyManagedKieServerConnectionIntegrationTest.java
+++ b/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/emptymanagedkieserver/EmptyManagedKieServerConnectionIntegrationTest.java
@@ -1,0 +1,173 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.cloud.integrationtests.architecture.emptymanagedkieserver;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Collection;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


mswiderski:
@sutaakar looks like we have reference implementation then, do you offer TCK too :)

framework-cloud/framework-cloud-common/src/main/java/org/kie/cloud/common/provider/KieServerClientProvider.java
```diff
@@ -33,6 +34,10 @@
     private static final long KIE_SERVER_TIMEOUT = 300_000L;
 
     public static KieServicesClient getKieServerClient(KieServerDeployment kieServerDeployment) {
+        return getKieServerClient(kieServerDeployment.getUrl(), kieServerDeployment);
+    }
+
+    public static KieServicesClient getKieServerClient(URL kieServerUrl, KieServerDeployment kieServerDeployment) {
```
jakubschwan: URL kieServerUrl is not used in this method. I think that this should be used in `KieServicesFactory.newRestConfiguration(kieServerUrl.toString(), kieServerDeployment.getUsername(), kieServerDeployment.getPassword(), KIE_SERVER_TIMEOUT);`

sutaakar: Good catch, will fix it.

test-cloud/pom.xml
```diff
@@ -128,6 +128,16 @@
                 </files>
               </configuration>
             </plugin>
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-failsafe-plugin</artifactId>
+              <configuration>
+                <excludes>
+                  <!-- Test use template which is available just for PostgreSQL -->
+                  <exclude>**/EmptyManagedKieServerConnectionIntegrationTest.java</exclude>
```
jiripetrlik: Maybe, we can consider to use junit categories for exclusions. I think that it will be quite painful to exclude test classes one by one.

sutaakar: Well, categories aren't as good solution either as they are specific for a test.
Maybe the way would be to have a check on scenario level, verifying that the scenario is available. @jiripetrlik What do you think?

jiripetrlik: What do you mean by "scenario level". How can it be implemented?

sutaakar: For example by adding method DeploymentScenario.isAvailable() . Will return true is the scenario has all its resources ready to be used.
Each test could have then assumeTrue(deploymentScenario.isAvailable()) .

jiripetrlik: Well, it sounds like a good solution.

sutaakar:
Test adjusted with latest cloud changes.

PR #58, opened 2017-10-09
Liveness probe.

jiripetrlik:
Tests for liveness probes in workbench and kie server.

### DIFF
--- a/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/deployment/CommandExecutionResult.java
+++ b/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/deployment/CommandExecutionResult.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.cloud.api.deployment;
+
+import java.io.ByteArrayOutputStream;
+
+public class CommandExecutionResult {
+    private ByteArrayOutputStream output;
+    private ByteArrayOutputStream error;
+
+    public String getOutput() {
+        return output.toString();
+    }
+
+    public void setOutput(ByteArrayOutputStream output) {
+        this.output = output;
+    }
+
+    public String getError() {
+        return error.toString();
+    }
+
+    public void setError(ByteArrayOutputStream error) {
+        this.error = error;
+    }
+}
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerInstanceImpl.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerInstanceImpl.java
@@ -15,6 +15,8 @@
 
 package org.kie.cloud.openshift.deployment;
 
+import static org.kie.cloud.openshift.util.CommandUtil.runCommandImpl;
+
 import java.io.BufferedOutputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
@@ -66,19 +68,7 @@ public class KieServerInstanceImpl implements KieServerInstance {
     }
 
     @Override public CommandExecutionResult runCommand(String... command) {
-        ByteArrayOutputStream output = new ByteArrayOutputStream();
-        ByteArrayOutputStream error = new ByteArrayOutputStream();
-
-        CommandExecutionResult commandExecutionResult = new CommandExecutionResult();
-        commandExecutionResult.setOutput(output);
-        commandExecutionResult.setError(error);
-
-        ExecWatch execWatch = openShiftController.getClient().pods().inNamespace(namespace).withName(podName)
-                .writingOutput(output)
-                .writingError(error)
-                .exec(command);
-
-        return commandExecutionResult;
+        return runCommandImpl(openShiftController.getClient().pods().inNamespace(namespace).withName(podName), command);
     }
 
     @Override public String getLogs() {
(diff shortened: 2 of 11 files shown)


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/probe/LivenessProbeIntegrationTest.java
```diff
@@ -0,0 +1,78 @@
...
+    @Test
+    public void testWorkbenchLivenessProbe() {
+        WorkbenchDeployment workbenchDeployment = deploymentScenario.getWorkbenchDeployment();
+        WorkbenchInstance workbenchInstance = (WorkbenchInstance) workbenchDeployment.getInstances().get(0);
+        String brokenPodName = workbenchInstance.getName();
+        logger.info("Running undepoloy command '{}' for workbench", UNDEPLOY_COMMAND);
+        workbenchInstance.runCommand("/bin/bash", "-c", UNDEPLOY_COMMAND);
+
+        logger.info("Waiting for liveness probe to kill workbench");
+        waitMilliseconds(KILL_POD_TIME);
```
sutaakar: Hardcoded thread sleep isn't a good approach as it wastes time. Could it be refactored to periodically check if pod is killed?

sutaakar:
A general idea (inspired by this PR) - do we need all those Instance implementations and interfaces? Things like SmartRouterInstance, KieServerInstance .... What does it bring to us?  Wouldn't it be simpler to have just one interface (Instance) and an implementation (OpenShiftInstance)?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerInstanceImpl.java
```diff
@@ -54,6 +65,22 @@ public void setPodName(String podName) {
         return podName;
     }
 
+    @Override public CommandExecutionResult runCommand(String... command) {
```
jakubschwan: I see that, runCommand method is same for all instances. Can we add implementation into some parent class e.g. OpenShiftInstance to avoid duplicate code?

jiripetrlik:
Fixed according to review.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/util/TimeUtils.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.function.BooleanSupplier;
+
+public class TimeUtils {
+
+    private static final long DEFAULT_WAIT_STEP = 5000;
+
+    public static void waitMilliseconds(long milliseconds) {
```
sutaakar: Just an idea, wouldn't it be more flexible to replace this by something like: wait(java.time.Duration duration) ? That way we would simplify its usage, to have wait(Duration.ofMinutes(1)) instead of waitMilliseconds(60000).

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/WorkbenchRuntimeInstanceImpl.java
```diff
@@ -32,6 +36,22 @@ public OpenShiftController getOpenShiftController() {
         return namespace;
     }
 
+    @Override public CommandExecutionResult runCommand(String... command) {
```
sutaakar: I guess this should be replaced as in all other InstanceImpls.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/util/TimeUtils.java
jiripetrlik: Yes, thank you. It's good idea.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/probe/LivenessProbeIntegrationTest.java
```diff
@@ -0,0 +1,86 @@
...
+    private static final Duration KILL_POD_TIME = Duration.ofSeconds(60);
+    private static final Logger logger = LoggerFactory.getLogger(LivenessProbeIntegrationTest.class);
+
+    @Test
+    public void testWorkbenchLivenessProbe() {
+        WorkbenchDeployment workbenchDeployment = deploymentScenario.getWorkbenchDeployment();
+        WorkbenchInstance workbenchInstance = (WorkbenchInstance) workbenchDeployment.getInstances().get(0);
+        String brokenPodName = workbenchInstance.getName();
+        logger.info("Running undepoloy command '{}' for workbench", UNDEPLOY_COMMAND);
+        workbenchInstance.runCommand("/bin/bash", "-c", UNDEPLOY_COMMAND);
```
jhrcek: I don't understand what's the purpose of returning CommandExecutionResult if you're not assigning it to any variable. Maybe it's not needed, but shouldn't you at least close the streams in it?

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/deployment/CommandExecutionResult.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+*/
+
+package org.kie.cloud.api.deployment;
+
+import java.io.ByteArrayOutputStream;
+
+public class CommandExecutionResult {
+    private ByteArrayOutputStream output;
```
jhrcek: Who's going to close those streams and when?

jhrcek:
@jiripetrlik you can ignore my comments if they don't make sense. I was just curious..

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/deployment/CommandExecutionResult.java
jiripetrlik: It is not necessary to close these streams because ByteArrayOutputStream does not consume any resources. It consumes only memory. Even close method is empty for these streams. It is also problem to find right place where to close them in this case.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/probe/LivenessProbeIntegrationTest.java
jiripetrlik: It's true. We don't need command output in this particular test. However, runCommand method is implemented in framework and possibly other tests will need such output.

PR #59, opened 2017-10-09
Template adjustments

sutaakar:
Replaced application template string by enum.
Skip tests in case template for defined database is missing.

### DIFF
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/OpenShiftTemplate.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/OpenShiftTemplate.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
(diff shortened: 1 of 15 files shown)


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/OpenShiftTemplate.java
```diff
@@ -0,0 +1,65 @@
...
+     * @return URL pointing to the template.
+     * @throws MissingResourceException If template is missing or template URL is malformed.
+     */
+    public URL getTemplateUrl() throws MissingResourceException {
+        String templateSystemPropertyValue = System.getProperty(templateSystemPropertyKey);
+
+        try {
+            return new URL(templateSystemPropertyValue);
+        } catch (MalformedURLException e) {
+            throw new MissingResourceException("Template referenced by system property " + templateSystemPropertyKey + " cannot be processed.", e);
```
jhrcek: Is the malformed url mentioned in the `e` when this happens? If so, nothing to do. Otherwise I'd include it in the exception message.

sutaakar: malformed URL added.

PR #69, opened 2017-10-20
Adjust external database tests to new template

jiripetrlik:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/builder/KieServerWithExternalDatabaseScenarioBuilderImpl.java
```diff
@@ -43,11 +44,15 @@ public KieServerWithExternalDatabaseScenarioBuilderImpl(OpenShiftController open
         this.envVariables = new HashMap<String, String>();
         this.envVariables.put(OpenShiftTemplateConstants.KIE_SERVER_USER, DeploymentConstants.getKieServerUser());
         this.envVariables.put(OpenShiftTemplateConstants.KIE_SERVER_PWD, DeploymentConstants.getKieServerPassword());
+        this.envVariables.put(OpenShiftTemplateConstants.KIE_SERVER_HOST, "aaa");
```
sutaakar: Is there any specific reason for this or just leftover?

jiripetrlik: This parameter is required by template. But maybe there is probably no reason for it. I've changed the code to be more clear.

PR #71, opened 2017-10-24
Raise version to 7.5.0-SNAPSHOT

sutaakar:
To be in sync with master.

### DIFF
--- a/framework-cloud/cloud-deployment-plugin/pom.xml
+++ b/framework-cloud/cloud-deployment-plugin/pom.xml
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.kie.cloud</groupId>
     <artifactId>framework-cloud</artifactId>
-    <version>7.4.1-SNAPSHOT</version>
+    <version>7.5.0-SNAPSHOT</version>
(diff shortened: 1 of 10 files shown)


framework-cloud/cloud-deployment-plugin/pom.xml
```diff
@@ -7,7 +7,7 @@
   <parent>
     <groupId>org.kie.cloud</groupId>
     <artifactId>framework-cloud</artifactId>
-    <version>7.4.1-SNAPSHOT</version>
+    <version>7.5.0-SNAPSHOT</version>
```
jhrcek: Is it really necessary to have this in all the poms? Wouldn't using `${project.parent.version}` be better?

sutaakar:
test this

sutaakar:
test this

framework-cloud/cloud-deployment-plugin/pom.xml
sutaakar: @jhrcek Unfortunately no, Maven isn't able to resolve expression in parent version.

PR #74, opened 2017-10-26
GUVNOR-3559: Verify router uses all available pods

jiripetrlik:


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
```diff
@@ -0,0 +1,92 @@
...
+        Assertions.assertThat(processDefinitions).isNotNull();
+        Assertions.assertThat(processDefinitions.stream().anyMatch(p -> p.getId().equals(LOG_PROCESS_ID)));
+
+        ProcessServicesClient processServicesClient = kieServerClientRouter.getServicesClient(ProcessServicesClient.class);
+        for (int i = 0; i < PROCESS_NUMBER; i++) {
+            processServicesClient.startProcess(CONTAINER_ID, LOG_PROCESS_ID);
+        }
+
+        List<Instance> kieServerInstances = deploymentScenario.getKieServerDeployment().getInstances();
+        Assertions.assertThat(kieServerInstances.size()).isGreaterThan(1);
```
sutaakar: I would say that there should be exactly 2 instances, so it should be checked exactly.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
```diff
@@ -0,0 +1,92 @@
...
+import org.kie.server.client.QueryServicesClient;
+import org.kie.server.integrationtests.controller.client.KieServerMgmtControllerClient;
+
+public class SmartRouterLoadBalancing extends AbstractCloudIntegrationTest<WorkbenchRuntimeSmartRouterKieServerDatabaseScenario> {
+
+    private KieServerMgmtControllerClient kieControllerClient;
+    private KieServicesClient kieServerClient;
+    private KieServicesClient kieServerClientRouter;
+
+    private static final int PROCESS_NUMBER = 100;
```
sutaakar: I would say that 10 should be enough, no need for 100. @jiripetrlik What do you think about that?

jhrcek: In my opinion running 100 or processes is relatively faster than the time it takes to deploy the scenario. On the other hand giving jBPM more to crunch might potentially better find some instabilities/race condition. So I think 100 is better in this case.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
```diff
@@ -0,0 +1,92 @@
...
+
+        ProcessServicesClient processServicesClient = kieServerClientRouter.getServicesClient(ProcessServicesClient.class);
+        for (int i = 0; i < PROCESS_NUMBER; i++) {
+            processServicesClient.startProcess(CONTAINER_ID, LOG_PROCESS_ID);
+        }
+
+        List<Instance> kieServerInstances = deploymentScenario.getKieServerDeployment().getInstances();
+        Assertions.assertThat(kieServerInstances.size()).isGreaterThan(1);
+        for (Instance kieServerInstance : kieServerInstances) {
+            Assertions.assertThat(kieServerInstance.getLogs()).contains(LOG_MESSAGE);
```
jhrcek: As far as load balancing goes, wouldn't it be better to check that the instances have been distributed more or less evenly (say by expecting that each instance has "at least 30%" of messages)?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    public void testRouterLoadBalancing() {
+        MavenDeployer.buildAndDeployMavenProject(ClassLoader.class.getResource("/kjars-sources/definition-project-snapshot").getFile());
+
+        kieControllerClient = KieServerControllerClientProvider.getKieServerMgmtControllerClient(deploymentScenario.getWorkbenchRuntimeDeployment());
+        kieServerClient = KieServerClientProvider.getKieServerClient(deploymentScenario.getKieServerDeployment());
+
+        KieServerInfo serverInfo = kieServerClient.getServerInfo().getResult();
+        kieControllerClient.saveContainerSpec(serverInfo.getServerId(), serverInfo.getName(), CONTAINER_ID, CONTAINER_ALIAS, PROJECT_GROUP_ID, DEFINITION_PROJECT_SNAPSHOT_NAME, DEFINITION_PROJECT_SNAPSHOT_VERSION, KieContainerStatus.STARTED);
```
jhrcek: Please reformat the code to avoid too-long lines. It's hard to read and when reviewing PRs on github it requires scrolling. I'd say that lines having 120+ characters (which is also the width of github's file preview) should be broken into multiple narrower lines.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
sutaakar: Right, good point.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
jiripetrlik: This check was removed.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
jiripetrlik: Good point @jhrcek . I will keep number of processes and check that at least one quarter was send to each execution server.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterLoadBalancing.java
jiripetrlik: Thanks, it was reformated.

jiripetrlik:
Thank you all for your reviews. Please see new version.

PR #84, opened 2017-11-03
Parallel test execution

sutaakar:


sutaakar:
@jiripetrlik I think that we could keep it on current level for now. If everything will go smoothly, without problems, we can consider raising thread number in future.

PR #93, opened 2017-11-16
Cloud runtime architectures tests

jakubschwan:


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
```diff
@@ -0,0 +1,291 @@
...
+
+/**
+ * Test for Architecture 3 from
+ * http://mswiderski.blogspot.cz/2017/08/cloud-runtime-architectures-for-jbpm.html
+ */
+public class EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest extends AbstractCloudArchitectureIntegrationTest<GenericScenario> {
+
+    private static final Logger logger = LoggerFactory.getLogger(EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.class);
+
+    private static final String RANDOM_URL_PART = "-" + UUID.randomUUID().toString().substring(0, 4);
```
sutaakar: This attribute shouldn't be static, otherwise components in different tests will share same URL.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
```diff
@@ -0,0 +1,291 @@
...
+        deployAndStartContainer(kieServerClientABC, CONTAINER_ID_ABC, deploymentScenario.getKieServerDeployments().get(0));
+        waitUntilKieServerLogsContain(deploymentScenario.getKieServerDeployments().get(0), "Container cont-id-abc (for release id org.kie.server.testing:definition-project-snapshot:1.0.0-SNAPSHOT) successfully started");
+        verifySmartRouterContainsKieServers(smartRouterAdminClient, 1, Arrays.asList(CONTAINER_ID_ABC), kieServerIdABC);
+        verifySmartRouterHostPerContainerContainsServers(smartRouterAdminClient, CONTAINER_ID_ABC, kieServerIdABC);
+        verifySmartRouterHostPerServerContainsContainers(smartRouterAdminClient, kieServerIdABC, CONTAINER_ID_ABC);
+
+        //check in smart router - kie servers and containers
+        verifySmartRouterContainsKieServers(smartRouterAdminClient, 1, Arrays.asList(CONTAINER_ID_ABC), kieServerIdABC);
+        verifySmartRouterHostPerContainerContainsServers(smartRouterAdminClient, CONTAINER_ID_ABC, kieServerIdABC);
+        verifySmartRouterHostPerServerContainsContainers(smartRouterAdminClient, kieServerIdABC, CONTAINER_ID_ABC);
```
sutaakar: These three lines seems to be a duplication of above ones.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
```diff
@@ -0,0 +1,291 @@
...
+        verifyServerTemplateContainsKieServers(kieControllerClient, kieServerIdABC, 3);
+        verifyServerTemplateContainsKieServers(kieControllerClient, kieServerIdDEF, 3);
+        verifyServerTemplateContainsKieServers(kieControllerClient, kieServerIdGHI, 3);
+        verifyServerTemplateContainsContainers(kieControllerClient, SMART_ROUTER_ID, CONTAINER_ID_ABC);
+        verifySmartRouterContainsKieServers(smartRouterAdminClient, 3, Arrays.asList(CONTAINER_ID_ABC), kieServerIdABC);
+    }
+
+    @Test
+    @Ignore("https://issues.jboss.org/browse/RHBPMS-4899 && https://issues.jboss.org/browse/JBPM-6623")
+    public void testWorkingWithSignalsAndTasksDeployments() {
```
sutaakar: I think it would be better to separate signals and tasks to separate test methods. When they are combined together here then it makes the test less readable.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyUnmanagedKieServersWithSmartRouterIntegrationTest.java
```diff
@@ -0,0 +1,242 @@
...
+        assertThat(smartQueryServicesClient.findProcessInstancesByStatus(Arrays.asList(org.kie.api.runtime.process.ProcessInstance.STATE_COMPLETED), 0, 10)).hasSize(2);
+
+        assertThat(smartQueryServicesClient.findProcessInstancesByStatus(Arrays.asList(org.kie.api.runtime.process.ProcessInstance.STATE_ACTIVE), 0, 10)).hasSize(4);
+        List<TaskSummary> activeTasks = smartTaskServicesClient.findTasks(USER_YODA, 0, 10);
+        assertThat(activeTasks).hasSize(1);
+        assertThat(activeTasks.get(0).getContainerId()).isEqualTo(CONTAINER_ID_DEF);
+        assertThat(activeTasks.get(0).getProcessInstanceId()).isEqualTo(signalProcessInstances.get(2));
+
+        smartTaskServicesClient.startTask(CONTAINER_ID_DEF, activeTasks.get(0).getId(), USER_YODA);
+        smartTaskServicesClient.completeTask(CONTAINER_ID_DEF, activeTasks.get(0).getId(), USER_YODA, Collections.EMPTY_MAP);
```
sutaakar: It is better to use Collections.emptyMap() instead of Collections.EMPTY_MAP for typed parameters  - to get rid of type safety warning.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/ManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
```diff
@@ -0,0 +1,293 @@
...
+        return signalProcessInstances;
+    }
+
+    protected int findServerDeploymentWithName(String kieServerName) {
+        for (int i = 0; i < deploymentScenario.getKieServerDeployments().size(); i++) {
+            if (deploymentScenario.getKieServerDeployments().get(i).getUrl().toString().contains(kieServerName)) {
+                return i;
+            }
+        }
+        return -1;
```
sutaakar: This seems to be invalid option which can cause IndexOutOfBoundsException. I think it would be better to fail fast - throwing some runtime exception with clear message.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
jakubschwan: Tests shouldn't share same URL, because tests from one class aren't parallel. But I agree, that static shouldn't be there.

sutaakar: Yes, for now :) It may change in future.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
```diff
@@ -0,0 +1,287 @@
...
+        assertThat(smartTaskServicesClient.findTasks(USER_YODA, 0, 10)).hasSize(0);
+    }
+
+    private void waitUntilKieServerLogsContain(KieServerDeployment kieServerDeployment, String logMessage) {
+        for (Instance kieServerInstance : kieServerDeployment.getInstances()) {
+            TimeUtils.wait(Duration.ofSeconds(15), Duration.ofSeconds(1), () -> kieServerInstance.getLogs().contains(logMessage));
+        }
+    }
+
+    private List<Long> createSignalProcesses() {
```
jhrcek: This could be made much more concise, e.g

```
private List<Long> createSignalProcesses() {
        // Start 2 instance of SignalUserTask each process on each container
        return Stream.of(CONTAINER_ID_ABC, CONTAINER_ID_ABC, CONTAINER_ID_DEF, CONTAINER_ID_DEF, CONTAINER_ID_GHI, CONTAINER_ID_GHI)
                .map(containerId -> smartProcessServicesClient.startProcess(containerId, SIGNALUSERTASK_PROCESS_ID))
                .collect(toList());
    }
```

The same applies to the remaining 4 instances of this helper method.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,184 @@
...
+import org.kie.server.controller.api.model.spec.ContainerSpec;
+import org.kie.server.integrationtests.controller.client.KieServerMgmtControllerClient;
+import org.kie.server.integrationtests.router.client.KieServerRouterClient;
+import org.kie.server.router.Configuration;
+
+public abstract class AbstractCloudArchitectureIntegrationTest extends AbstractCloudIntegrationTest<GenericScenario> {
+
+    protected static final String SMART_ROUTER_ID = "test-kie-router";
+
+    protected static final String CONTAINER_ID_ABC = CONTAINER_ID + "-abc";
```
jhrcek: You can save several hundred characters using shared type declaration for multiple constants of the same type, like
```
protected static final String
  A = "a",
  B = "b",
  C = "c";
```

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/UnmanagedKieServersWithSmartRouterIntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private static final Logger logger = LoggerFactory.getLogger(UnmanagedKieServersWithSmartRouterIntegrationTest.class);
+
+    private final String RANDOM_URL_PREFIX = UUID.randomUUID().toString().substring(0, 4) + "-";
+
+    private static final String SMART_ROUTER_HOSTNAME = SMART_ROUTER_NAME + DeploymentConstants.getDefaultDomainSuffix();
+
+    private DeploymentSettings smartRouter;
+
+    private DeploymentSettings kieServerABC;
```
jhrcek: (stricly optional) the same comment as above and throughout this PR for repeated type declarations. I find 
```
private DeploymentSettings
    kieServerABC,
    kieServerDEF,
    kieServerGHI;
```
takes some unnecessary verbosity away and makes it easier on the eyes, but stick with your version if you like it better.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
jakubschwan: I change this in unmanaged architectures. In managed mode are processes started throw Smart router and in Kie server.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/EmptyManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
jakubschwan: If I separate test method, then test will take 5min longer. This time is consumed by OpenShift configuration, the most is taken by S2I builds.
I think it should be better to save some time here and separate all test to private methods and run only one test per class.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+    protected String repositoryName;
+
+    protected String getKieServerId(KieServicesClient kieServerClient) {
+        ServiceResponse<KieServerInfo> serverInfo = kieServerClient.getServerInfo();
+        assertThat(serverInfo.getType()).isEqualTo(ServiceResponse.ResponseType.SUCCESS);
+
+        return serverInfo.getResult().getServerId();
+    }
+
+    protected void verifyClientKieServersLocationCount(KieServicesClient kieServerClient, final int numberOfExpectedServers) {
```
triceo: This method is actually two methods. Let's have it be two real methods.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    protected void verifyContainerIsDeployed(KieServicesClient kieServerClient, String containerId) {
+        ServiceResponse<KieContainerResourceList> containers = kieServerClient.listContainers();
+        assertThat(containers.getType()).isEqualTo(ServiceResponse.ResponseType.SUCCESS);
+
+        List<KieContainerResource> containerList = containers.getResult().getContainers();
+        assertThat(containerList).hasSize(1);
+        assertThat(containerList.get(0).getContainerId()).isEqualTo(containerId);
```
triceo: Personally, I'd just put the `cointanerList.get(0)` into its own variable.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+    protected void verifyServerTemplateContainsContainers(KieServerMgmtControllerClient kieControllerClient, String serverTemplate, String... containerIds) {
+        Collection<ContainerSpec> containersSpec = kieControllerClient.getServerTemplate(serverTemplate).getContainersSpec();
+        assertThat(containersSpec).hasSize(containerIds.length);
+        List<String> containersSpecIds = getAllContainersSpecIds(containersSpec);
+        assertThat(containersSpecIds).containsExactlyInAnyOrder(containerIds);
+    }
+
+    private List<String> getAllContainersSpecIds(Collection<ContainerSpec> containerSpecs) {
+        List<String> ids = new ArrayList<>();
```
triceo: `return containerSpecs.stream().map(spec -> spec.getId()).collect(Collectors.toList());`

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+        return ids;
+    }
+
+    protected void verifyServerTemplateContainsKieServers(KieServerMgmtControllerClient kieControllerClient, String serverTemplate, int numberOfKieServers) {
+        Collection<ServerInstanceKey> kieServers = kieControllerClient.getServerTemplate(serverTemplate).getServerInstanceKeys();
+        assertThat(kieServers).hasSize(numberOfKieServers);
+    }
+
+    protected void verifySmartRouterContainsKieServers(KieServerRouterClient smartRouterAdminClient, int numberOfKieServers, List<String> containerIds, String... kieServerIds) {
```
triceo: This is a good use case for `SoftAssertions`.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+    protected void verifySmartRouterHostPerContainerContainsServers(KieServerRouterClient smartRouterAdminClient, String containerId, String... kieServerIds) {
+        Configuration routerConfig = smartRouterAdminClient.getRouterConfig();
+        assertThat(routerConfig.getHostsPerContainer().get(containerId)).contains(kieServerIds);
+    }
+
+    protected void scaleKieServerTo(int count, KieServerDeployment... deployments) {
+        scaleKieServerTo(count, Arrays.asList(deployments));
+    }
+
+    protected void scaleKieServerTo(int count, List<KieServerDeployment> deployments) {
```
triceo: Any reason for this method (and couple others) to not be `static`? What inner attributes of the class does it use or need?
Also, the order of its argument should, IMHO, be inverted. First say which deployments, then say how many.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+        for (Deployment deployment : deployments) {
+            deployment.scale(count);
+            deployment.waitForScale();
+        }
+    }
+
+    protected void assertActiveProcessInstanceCount(int expectedCount, QueryServicesClient... clients) {
```
triceo: Again, `static`. Also, it's a method whose only purpose is to iterate over the argument. Why not have the method assert on the client directly, and iterate on the outside?

I can see some value here, but in order for that value to be realized, the method would have to use `SoftAssertions`.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+        for (QueryServicesClient client : clients) {
+            assertThat(client.findProcessInstances(0, 10)).hasSize(expectedCount);
+        }
+    }
+
+    protected int findServerDeploymentWithName(String kieServerName) {
+        for (int i = 0; i < deploymentScenario.getKieServerDeployments().size(); i++) {
```
triceo: `.stream().filter().findFirst().orElseThrow()`

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+        assertThat(serverInfo.getType()).isEqualTo(ServiceResponse.ResponseType.SUCCESS);
+
+        return serverInfo.getResult().getServerId();
+    }
+
+    protected void verifyClientKieServersLocationCount(KieServicesClient kieServerClient, final int numberOfExpectedServers) {
+        if (numberOfExpectedServers == 0) {
+            ServiceResponse<KieServerInfo> serverInfo;
+            try {
+                serverInfo = kieServerClient.getServerInfo();
```
triceo: Personally, I'd fit these assertions into the try-catch block. Perhaps to its own method, `assertOnKieServerInfo()`.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+            } catch (KieServicesHttpException ex) {
+                //ok
+                assertThat(ex.getHttpCode()).isEqualTo(503);
+                return;
+            }
+            assertThat(serverInfo.getType()).isEqualTo(ServiceResponse.ResponseType.FAILURE);
+            assertThat(serverInfo.getMsg()).contains("404");
+        } else {
+            Set<String> locations = new HashSet<>();
+            for (int i = 0; i < numberOfExpectedServers * 3; i++) {
```
triceo: Why is the `i` even here? Never used anywhere. Does that mean that `kieServerClient.getServerInfo()` returns a different result every time?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
```diff
@@ -0,0 +1,188 @@
...
+                ServiceResponse<KieServerInfo> serverInfo = kieServerClient.getServerInfo();
+                assertThat(serverInfo.getType()).isEqualTo(ServiceResponse.ResponseType.SUCCESS);
+
+                locations.add(serverInfo.getResult().getLocation());
```
triceo: Another good case for `SoftAssertions`. Let it all evaluate, report all at the same time.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
jakubschwan: Yes, client return server info from different Kie server instances. This request can be affected by OpenShift router or Kie server load balancing, so we also repeat this request several time to be sure we have all.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
jakubschwan: I don't think so. I need to know here, if Service respond was success, so then I can get service response result without  NPE.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
triceo: This is exactly the kind of thing that needs to be written there as a comment. Otherwise the code hardly makes any sense.

Also... is this really the only way to test for this? Call this N times and hope that it returns what we want?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
triceo: So the result of the method depends on whether or not the assertion fails? Sorry, that's borderline evil. :-)

My suggestion: do whatever you need in the method, return the locations, and then assert on them.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/AbstractCloudArchitectureIntegrationTest.java
jakubschwan: This is only way I know.
We can try to get instances from Kie server deployment (OpenShift Pod representation in tests). But we are not able to say, if instances are correctly responding Kie servers.

PR #94, opened 2017-11-16
JBPM-6606: Verify Smart router balancing for container alias

sutaakar:


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/SmartRouterUpdateRedirection.java
```diff
@@ -110,18 +111,36 @@ public void testRouterLoadBalancing() {
...
+    public void testRouterContainerAliasLoadBalancing() {
+        deployProject(kieServerClient1, CONTAINER_ID, CONTAINER_ALIAS, Kjar.DEFINITION_SNAPSHOT);
+        deployProject(kieServerClient2, CONTAINER_ID_UPDATED, CONTAINER_ALIAS, Kjar.DEFINITION_SNAPSHOT);
+
+        for (int i = 0; i < RETRIES_NUMBER; i++) {
+            verifyProcessAvailableInContainer(smartRouterClient, CONTAINER_ALIAS, LOG_PROCESS_ID);
+        }
+
+        assertThat(kieServerDeployment1.getInstances()).hasSize(1);
+        assertThat(kieServerDeployment1.getInstances().get(0).getLogs()).contains(LOG_MESSAGE);
```
jakubschwan: This server, with CONTAINER_ID, shouldn't contains log messages. CONTAINER_ALIAS in the startProcess should start processes from the newest Kie container (CONTAINER_ID_UPDATED).

sutaakar: @jakubschwan Well, that is the question. Problem is that some requests (for example user task requests) need to be delivered to specific Kie server so it data objects can be correctly marshalled/unmarshalled .
Current implementation is that requests for container alias are spread to all Kie servers which have container with such alias. That may look inconsistent in some cases like long running rolling update.
@mswiderski Wouldn't it make more sense for Smart router to try to redirect requests with alias to Kie server with latest container id? It needs to be a user responsibility anyway to make sure that Kie server with latest container id is able to handle requests with previous container ids if there are still old processes available.

mswiderski: there is no way for router to know what is the latest container id. that's why we have alias which then pushes this responsibility to kie server that has enough information (like underlying db, deployed containers etc) to figure this out.

sutaakar: Well, problem is that Smart router has to decide where it will redirect the request, what Kie server should be able to satisfy it. But at the same time it doesn't have all needed informations to correctly decide it for all cases.
In case all Kie servers referenced by specific container alias will contain all containers associated with that specific alias then there would be no problem, as it will mean that any of those Kie servers will be able to handle any request. But I am afraid that it won't always be the case. I think that we should have some strategy to handle situations when Kie server1 will contain just container1 with alias cont-alias and Kie server2 will contain container1 and container2, both with alias cont-alias.
About latest container id, Smart router could be able to determine latest release id from container alias. Router's configuration contains map mapping container id/alias to ContainerInfo, which contains informations about release id.

mswiderski: @sutaakar that's the main requirement to use router with aliases. you cannot have different instances of kie server having different container ids under same alias. That simply won't work. Moreover, if you would allow it, that will open up for situation where people will start using different dbs etc. 

So recap on aliases, there is nothing preventing you from using different GAVs under same alias so in general there will be no clear way of saying which version is latest and which version has given process instance. And for that reason there is this strict requirement.

sutaakar:
Test adjusted according to comments to simulate supported use case.

PR #97, opened 2017-11-27
Move Util classes to framework-cloud-common

MarianMacik:
@sutaakar @jiripetrlik @jhrcek


framework-cloud/framework-cloud-common/src/main/java/org/kie/cloud/common/client/util/TimeUtils.java
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
 */
 
-package org.kie.cloud.integrationtests.util;
+package org.kie.cloud.common.client.util;
```
sutaakar: Just a suggestion - TimeUtils could be moved to different package as it is used also outside of client util classes.

MarianMacik: Moved to org.kie.cloud.common.time package. Took inspiration from org.kie.cloud.common.logs which contains InstanceLogUtil.

MarianMacik:
jenkins retest this

PR #108, opened 2018-01-03
Replace OpenShift client dependency by XTF

sutaakar:
XTF is placed in framework-common because it will replace git and maven modules in future.


sutaakar:
Do not merge yet, need to add repositories to proxy server.

sutaakar:
Jenkins retest this

PR #109, opened 2018-01-03
Minor usability enhancements

sutaakar:
Template is commented out to prevent test failures.
Deployment lists are cleaned to allow reusing of generic scenario instance between several test runs - needed for parameterized tests.


test-cloud/src/main/resources/templates-general.properties
```diff
@@ -4,5 +4,5 @@ kie.app.template.workbench=https://raw.githubusercontent.com/jboss-openshift/app
 kie.app.template.workbench-monitoring=https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-businesscentral-monitoring.json
 kie.app.template.workbench-monitoring.smartrouter=https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-businesscentral-monitoring-with-smartrouter.json
 kie.app.template.kie-server.database.external=https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-executionserver-externaldb.json
-kie.app.template.smartrouter=https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-smartrouter.json
+#kie.app.template.smartrouter=https://raw.githubusercontent.com/jboss-openshift/application-templates/bpmsuite-wip/bpmsuite/bpmsuite70-smartrouter.json
```
jhrcek: (optional) Maybe comment why this is disabled and when can it be reenabled?

sutaakar: Good point, comment added.

PR #112, opened 2018-01-05
drools/jbpm separation and minor updates

sutaakar:
Templates are separated now, please use drools profile for drools templates, jbpm + "some DB" profiles for jbpm templates.


test-cloud/pom.xml
```diff
@@ -141,8 +173,8 @@
               <groupId>org.codehaus.mojo</groupId>
               <artifactId>properties-maven-plugin</artifactId>
               <configuration>
-                <files>
```
jhrcek: I'm not 100% sure, but this "append" behaviour [should be default](http://blog.sonatype.com/2011/01/maven-how-to-merging-plugin-configuration-in-complex-projects/), so `combine.children="append"` should not be needed. Or it doesn't work that way and I need to update my model of the world :-)

sutaakar: @jhrcek It isn't default, default value is combine.children="merge"
With combine.children="merge" the configuration file specified in latter profile (for example if running with -Popenshift,jbpm,mysql) overrides configuration file from former profile.

jhrcek: Ah, so I got it wrong. Thanks

PR #116, opened 2018-01-11
Switch framework-maven to use XTF

sutaakar:


framework-cloud/framework-maven/src/main/java/org/kie/cloud/maven/MavenDeployer.java
```diff
@@ -52,33 +54,13 @@ public static void buildAndDeployMavenProject(String basedir) {
...
-            mvnArgs.add("-s");
-            mvnArgs.add(buildSettings);
-        }
-
-        MavenCli cli = new MavenCli();
-        int mvnRunResult = cli.doMain(mvnArgs.toArray(new String[mvnArgs.size()]), basedir, System.out, System.err);
-
-        Thread.currentThread().setContextClassLoader(classLoaderBak);
+        try {
+            MavenUtil mavenUtil = MavenUtil.forProject(Paths.get(basedir), Paths.get(SETTINGS_XML_PATH)).forkJvm();
```
jhrcek: Will this work if someone forgets to set `kjars.build.settings.xml` system property? (maybe it's handled already in the MavenUtil. If not, perhaps check that that file exists before calling MavenUtil..

sutaakar: Added handling of settings.xml file. It doesn't need to be defined (in case when user just want to install project or deploy project to the repo without authentication), but if it is defined then it has to point to existing location.

sutaakar:
@jiripetrlik Currently yes though GitHub doesn't seem to be used. But when we switch to XTF then GitHub support will most likely end.

PR #135, opened 2018-01-31
Adjust parameterized test classes to get readable output messages 

jakubschwan:


sutaakar:
Just an idea - if we will specify @Parameters(name = "{0}") then we don't have to implement toString methods for all of the scenarios. From my point of view they currently don't bring much informations.
@jakubschwan What do you think about it?

jakubschwan:
Removed toString methods and resolved conflicts

sutaakar:
@jakubschwan Can you please also delete {1} from parameters so just the test descriptions of scenarios are used?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/architecture/ManagedKieServersWithSmartRouterAndControllerIntegrationTest.java
```diff
@@ -59,10 +59,12 @@
 @RunWith(Parameterized.class)
 public class ManagedKieServersWithSmartRouterAndControllerIntegrationTest extends AbstractCloudArchitectureIntegrationTest {
 
-    @Parameter
+    @Parameter(value = 0)
+    public String testScenarioIndex;
```
jhrcek: (optional) confusing variable name - this looks more like testScenarioName ?

jakubschwan:
I think that this is not a good idea. We can't have scenario name for template, because we can use same scenario with some changes for same test case - e.g. WB+KS with or without external repo. We should set this name through method setName() and for scenarios there are builders and these builders configure environment variables. Combine these variables with just naming of the template can be confusing.
Also we parameterize settings builders (for S2I Kie servers)

jhrcek:
@jakubschwan ok, taking that back.

triceo:
FYI changes like this should be approached with caution - changing the FQN of a test (which this will likely do) will create a new test case in Polarion.

I suggest you think of that before you introduce a test case. Since otherwise you're just creating extra work for yourselves.

PR #143, opened 2018-02-08
Check node name for the pod

sutaakar:
For analysis

### DIFF
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/GenericScenarioImpl.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/GenericScenarioImpl.java
@@ -117,6 +117,8 @@ public class GenericScenarioImpl extends OpenShiftScenario implements GenericSce
         for (KieServerDeployment kieServerDeployment : kieServerDeployments) {
             kieServerDeployment.waitForScale();
         }
+
+        logNodeNameOfAllInstances();
     }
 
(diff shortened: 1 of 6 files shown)


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/GenericScenarioImpl.java
```diff
@@ -112,10 +113,20 @@ public void deploy() {
         logger.info("Waiting for Workbench deployment to become ready.");
         for (WorkbenchDeployment workbenchDeployment : workbenchDeployments) {
             workbenchDeployment.waitForScale();
+
+            for (Instance workbenchInstance : workbenchDeployment.getInstances()) {
```
jhrcek: (Strictly optional) This looks kind of repetitive. Personally I'd create a utility method
```
public void logNodeNameOfAllInstances(Deployment deployment, String whatIsIt) {
     deployment.getInstances().forEach(
         instance -> {
            String instanceNodeName = project.getOpenShiftUtil().getPod(instance.getName()).getSpec().getNodeName();
            logger.info("{} name: {} ", whatIsIt, instanceNodeName);
         }
     );
```

and then call it like

`logNodeNameOfAllInstances("Workbench", workbenchDeployment);`

sutaakar: Adjusted to remove duplicities

PR #165, opened 2018-04-09
Adjustments for rhpam70-authoring template

sutaakar:


sutaakar:
Failures are caused by another change in different template, will be addressed in next PR.

PR #171, opened 2018-04-16
Adjustments for rhpam70-sit template

sutaakar:
Can be merged when the template itself is merged and URL is adjusted.


framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/builder/WorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioBuilder.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param smartRouterId Smart router id.
+     * @return Builder with configured Smart router id.
+     */
+    WorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioBuilder withSmartRouterId(String smartRouterId);
+
+    /**
+     * @param kieServerId kie-server id
+     * @return Builder with kie-server id set
+     */
+    WorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioBuilder withKieServerId(String kieServerId);
```
jakubschwan: We should be able to configure ID for both Kie Servers in template, right? Or is there only one KieServerId with suffix for each deployment? Please change this to 'withKieServerOneId' and add 'withKieServerTwoId'.

sutaakar: Good catch, currently there is no way to set Kie server id of any Kie server in this template so I remove it. Question is if there should be a way to set it. I will ask in the template PR.

PR #178, opened 2018-04-27
Load template URLs on OpenShiftTemplate init, not in tests

jhrcek:
I'm trying to solve the problem with having to duplicate declarations in bxms-qe-tests for these often changing properties.

Advantages:
- It solves my problem, because properties are loaded within the kie-wb-cloud framework classes
- It removes some property declaration overhead in pom.xml and copying properties to failsafe plugin.
- The properties are defined in the same module / package as the enum representing the Templates

Disadvantages: 
- TestInfoPrinter prints `<value> = null` for these url properties, because it's looking up the values from system properties.
- The logic to load properties is implemented in procedural code, and is no longer declarative in pom.xml

Further ideas to think about: do we even need this level of indirection (loading properties from files)? Why not declare the URLs for templates directly in the enum constants? If you need to test with different template url locally, you can change the url in the enum. If you need to test different template url on jenkins, you can change the enum again, push to separate branch and rerun the job with that branch. Is the current approach (via prop. files loaded by property-maven-plugin) more flexible in some ways?


sutaakar:
@jhrcek Good points. What it would mean if we will define URLs directly in enums? Will it mean that there will be distinct enums for different databases? Or for drools and jbpm?
Or would it have sense to split scenarios themselves to drools/jbpm/db specific ones? So the tests will contain different lists of scenarios for drools and jbpm?

jhrcek:
@sutaakar In my opinion the Template enum should be dumb and each constant should correspond to 1 template URL. There should be layer of logic above that (e.g. `TemplateUtil`) which would select appropriate enum constant based on some external configuration(product / database etc). I'll implement this and add it in separate commit for you to check..

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/OpenShiftTemplatePropertiesLoader.java
```diff
@@ -0,0 +1,52 @@
+package org.kie.cloud.openshift.template;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Properties;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class OpenShiftTemplatePropertiesLoader {
```
triceo: Not sure what it is about this class, but I don't like it. Perhaps it's the fact that all methods are `static`, yet it is not a singleton? Perhaps it's the way how `static load()` relies on side-effects of `static loadTemplate...()`?

The entire thing could essentially just be a lazy-loaded field.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/TemplateSelector.java
```diff
@@ -0,0 +1,36 @@
+package org.kie.cloud.openshift.template;
+
+import java.util.Arrays;
+
+class TemplateSelector {
+
+    enum Database {
+        postgresql,
```
triceo: Maybe I'm wrong, but doesn't the code style dictate that enum values are all upper-case?

framework-cloud/framework-openshift/pom.xml
```diff
@@ -57,6 +57,11 @@
       <artifactId>commons-io</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
```
triceo: Personally, I'd say that including the entire library just to use a simple `StringUtils.isEmpty()` is overkill.

jhrcek: I agree. I wanted to use it for something else, but in the end I didn't. Removed

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/OpenShiftTemplatePropertiesLoader.java
jhrcek: > The entire thing could essentially just be a lazy-loaded field.

Not sure what you mean, but I did some simplification and made the constructor private..

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/TemplateSelector.java
jhrcek: Changed the names to all uppercase. I just wanted to avoid case conversion when comparing with potential string values. But in the end I guess that `equalsIgnoringCase` doesn't hurt..

jhrcek:
@sutaakar @jiripetrlik If you agree with these changes, I'll create a PR to simplify jenkins job configuration tomorrow.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/template/TemplateSelector.java
```diff
@@ -0,0 +1,36 @@
...
+
+class TemplateSelector {
+
+    enum Database {
+        POSTGRESQL,
+        MYSQL,
+        GENERAL
+    }
+
+    enum Product {
```
triceo: @jhrcek One more thing crossed my mind - Is `Product` the proper name for something which has values of `DROOLS` and `JBPM`?

jhrcek: Good point changed to Project in last commit

PR #181, opened 2018-05-02
RHPAM-802: HTTPS secret is a required parameter

sutaakar:


sutaakar:
Test failures are caused by other changes, will be addressed in next PR.

PR #182, opened 2018-05-02
Retrieve route URI using route target service

sutaakar:
This should fix the remaining test issues related to recent template change. There are also bugs directly in templates, will be fixed by appropriate PR.


sutaakar:
Errors should be fixed by https://github.com/jboss-container-images/rhpam-7-openshift-image/pull/53

PR #184, opened 2018-05-22
Add tests for rhpam70-authoring-ha template

sutaakar:


framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/DeploymentScenarioBuilderFactory.java
```diff
@@ -35,6 +36,7 @@
     WorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioBuilder getWorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioBuilder();
     KieServerWithExternalDatabaseScenarioBuilder getKieServerWithExternalDatabaseScenarioBuilder();
     KieServerWithDatabaseScenarioBuilder getKieServerWithDatabaseScenarioBuilder();
+    ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder getClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder();
```
jhrcek: Just to clarify: what does the work `Persistent` in scenario mean? Doesn't presence of `Database` already imply persistence?

sutaakar: @jhrcek Not necessarily. Persistent means that the deployment state is persisted through pod restarts. Theoretically database doesn't imply persistence as there may be ephemeral databases (though not used in our templates).

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/ClusteredWorkbenchKieServerDatabasePersistentScenarioImpl.java
```diff
@@ -0,0 +1,120 @@
...
+        super.deploy();
+
+        logger.info("Processing template and creating resources from " + OpenShiftTemplate.CLUSTERED_WORKBENCH_KIE_SERVER_DATABASE_PERSISTENT.getTemplateUrl().toString());
+        envVariables.put(OpenShiftTemplateConstants.IMAGE_STREAM_NAMESPACE, project.getName());
+        project.processTemplateAndCreateResources(OpenShiftTemplate.CLUSTERED_WORKBENCH_KIE_SERVER_DATABASE_PERSISTENT.getTemplateUrl(), envVariables);
+
+        workbenchDeployment = new WorkbenchDeploymentImpl(project);
+        workbenchDeployment.setUsername(DeploymentConstants.getWorkbenchUser());
+        workbenchDeployment.setPassword(DeploymentConstants.getWorkbenchPassword());
+        workbenchDeployment.scale(1);
```
jakubschwan: If workbench is clustered, shouldn't we want default count of replicas > 1 ?

sutaakar: I set it to one to keep the same approach in all scenarios. That way all the tests can expect the same configuration available.

jakubschwan: Ok, make sense

sutaakar:
Test failure is related to image version

PR #185, opened 2018-05-23
Enable and update timer test

sutaakar:
Process updated due to synchronization issues. Now the process contains intermediate timer event generating 3 events, those events trigger PrintingNode.


sutaakar:
Test failures unrelated to the changes.

PR #187, opened 2018-05-25
Log failures when instance log is stored

sutaakar:
Log error instead of throwing exception so the test can continue.
Original approach prevented project to be deleted in case of any exception when retrieving logs.


sutaakar:
Failures are unrelated.

PR #188, opened 2018-06-04
Verify consistency of parameters in available templates

sutaakar:


test-cloud/test-cloud-template/src/test/java/org/kie/cloud/tests/template/TemplateParameterTest.java
```diff
@@ -0,0 +1,97 @@
...
+        OpenShiftConfig openShiftConfig = new OpenShiftConfigBuilder()
+                .withDisableApiGroupCheck(true)
+                .build();
+
+        openShiftClient = new DefaultOpenShiftClient(openShiftConfig);
+    }
+
+    @Test
+    public void testTemplateParameterDuplicities() {
+        for (OpenShiftTemplate openShiftTemplate : availableOpenShiftTemplates) {
```
jhrcek: I like this PR!  But instead of using for loop in these tests, we should be probably using `@Parametrized` test. In that way if there is problem introduced in 2 templates, we'll see separate failure for each of them, while this approach will only show one failure. WDYT?

sutaakar:
@jhrcek Updated according to your suggestion.

PR #191, opened 2018-06-13
Fix ReadinessProbeIntegrationTest

sutaakar:
It seems that Workbench client doesn't handle pod restarts well. Fixed by creating of new client to work with restarted pod.


sutaakar:
retest this please

sutaakar:
retest this please

PR #195, opened 2018-06-25
Implement support for Gogs

jiripetrlik:


framework-cloud/framework-git/src/main/java/org/kie/cloud/git/gogs/GogsGitProvider.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.http.entity.ContentType;
+import org.kie.cloud.git.AbstractGitProvider;
+import org.apache.http.client.fluent.Request;
+import org.kie.cloud.git.constants.GitConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GogsGitProvider extends AbstractGitProvider {
+
+    private String url;
```
jhrcek: Can we make these final, please?

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/gogs/GogsGitProvider.java
```diff
@@ -0,0 +1,93 @@
...
+        createRepository(repositoryName);
+        pushToGitRepository(getRepositoryUrl(repositoryName), repositoryPath,
+                user, password);
+
+        return repositoryName;
+    }
+
+    @Override public void deleteGitRepository(String repositoryName) {
+        try {
+            final int statusCode = Request.Delete(url + URL_API_SUFFIX + URL_REPOSITORY_SUFFIX + "/" + user + "/" +
```
jhrcek: Building URLs using string concatenation can be error prone (e.g. what if some piece ends/doesn't end with '/'?). Unfortunately I'm not aware of java ecosystem that solves this in a nicer way :P

pom.xml
```diff
@@ -160,6 +177,12 @@
         <artifactId>maven-verifier</artifactId>
         <version>${version.org.apache.maven.shared}</version>
       </dependency>
+
+      <dependency>
+        <groupId>org.eclipse.jgit</groupId>
+        <artifactId>org.eclipse.jgit</artifactId>
+        <version>4.11.0.201803080745-r</version>
```
sutaakar: Just a nitpick, please put the version into the properties so they all are in one place.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/AbstractCloudIntegrationTest.java
```diff
@@ -98,10 +98,6 @@ public void initializeDeployment() {
     public void cleanEnvironment() {
         if (deploymentScenario != null) {
             deploymentScenario.undeploy();
-
-            if (gitProvider != null) {
-                gitProvider.deleteGitRepository(deploymentScenario.getNamespace());
```
sutaakar: Why did you delete this code? Does it cause any issue with Gogs?

jiripetrlik: Yes, it seems that delete method is called twice. At least in some cases. For example in "KieServerS2iJbpmIntegrationTest". It is called here:
- https://github.com/kiegroup/kie-cloud-tests/blob/master/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmIntegrationTest.java#L103
- https://github.com/kiegroup/kie-cloud-tests/blob/master/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/AbstractCloudIntegrationTest.java#L103

pom.xml
jiripetrlik: Good catch, will improve it.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/AbstractCloudIntegrationTest.java
sutaakar: Ok, but why did you delete it here?

jiripetrlik: It's also problem that in this place it is trying to delete wrong git project. Will try to get to know more and decide better. :-)

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/gogs/GogsGitProvider.java
```diff
@@ -0,0 +1,93 @@
...
+    private String user;
+    private String password;
+
+    private static final String URL_API_SUFFIX = "/api/v1";
+    private static final String URL_CREATE_REPOSITORY_SUFFIX = "/user/repos";
+    private static final String URL_REPOSITORY_SUFFIX = "/repos";
+
+    private static final Logger logger = LoggerFactory.getLogger(GogsGitProvider.class);
+
+    @Override public String createGitRepositoryWithPrefix(String repositoryPrefixName, String repositoryPath) {
```
jakubschwan: Just a nitpick, from my point of view for reading purpose is better to have Override annotation on the separate line.

README.md
```diff
@@ -49,6 +49,14 @@ Can be found in framework-git, class org.kie.cloud.git.constants.GitConstants
 | github.username     |        | Username for logging into GitHub      |
 | github.password     |        | Password for logging into GitHub      |
 
+#### Gogs
+
+| \<specific-params\> | Value  |  Meaning                              |
+| ------------------- | ------ | ------------------------------------- |
+| git.provider        | Gogs   |                                       |
+| gogs.username       |        | Username for logging into Gogs        |
+| gogs.password       |        | Password for logging into GitHub      |
```
winklerm: Github or Gogs?

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/GitProviderService.java
```diff
@@ -0,0 +1,62 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
winklerm: 2018?

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/GitProviderService.java
```diff
@@ -0,0 +1,62 @@
...
+import org.kie.cloud.git.github.GitHubGitProviderFactory;
+import org.kie.cloud.git.gitlab.GitLabGitProviderFactory;
+import org.kie.cloud.git.gogs.GogsGitProviderFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class GitProviderService {
+    private final Map<String, GitProviderFactory> providerFactories;
+
+    private static final Logger logger = LoggerFactory.getLogger(GitProviderService.class);
```
winklerm: I would probably suggest LOGGER, since it is a private static final field. But I am not aware of code conventions in this particular repository.

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/gogs/GogsGitProvider.java
```diff
@@ -0,0 +1,147 @@
...
+    private final String url;
+    private final String user;
+    private final String password;
+
+    private static final String URL_API_SUFFIX = "/api/v1/";
+    private static final String URL_CREATE_REPOSITORY_SUFFIX = "user/repos";
+    private static final String URL_REPOSITORY_SUFFIX = "repos/";
+    private static final String URL_GIT_SUFFIX = ".git";
+
+    private static final Logger logger = LoggerFactory.getLogger(GogsGitProvider.class);
```
winklerm: The same as before, LOGGER?

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/constants/GitConstants.java
```diff
@@ -52,6 +60,27 @@ public static String getGitHubPassword() {
...
+
+    public static String getGogsUser() {
+        return System.getProperty(GOGS_USER);
+    }
+
+    public static String getGogsPassword() {
+        return System.getProperty(GOGS_PASSWORD);
+    }
+
+    public static String readMandatoryParameter(String value, String parameterName) {
```
jhrcek: Wouldn't it be better to make this implementation detail of each of the methods? I don't like that the API doesn't enforce the "mandatoriness". As it is you can misuse it by calling `GitConstants.getGogsPassword`and it would fail if not seat. How about this instead?
```
public static String getGogsPassword() {
       return readMandatoryParameter(GOGS_PASSWORD);
}

private static String readMandatoryParameter(String parameterName) {
        Strinv value = System.getProperty(parameterName)
        if (value == null) {
            logger.error("Parameter {} must be specified", parameterName);
            throw new RuntimeException("Parameter " + parameterName + " must be specified");
        }
        return value;
 }
```

README.md
jiripetrlik: Good catch. Fixed.

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/GitProviderService.java
jiripetrlik: Thanks. :-) Fixed.

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/constants/GitConstants.java
jiripetrlik: @jhrcek Yes, you are right. But it is problem of all parameters in Kie cloud tests. I think we can polish it in another PR.

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/gogs/GogsGitProvider.java
jiripetrlik: We use logger with small letters everywhere in this test suite. I will keep it this way.

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/GitProviderService.java
jiripetrlik: We use logger with small letters everywhere in this test suite. I will keep it this way.

jiripetrlik:
@winklerm Thank you for valuable comments. Can you please approve now?

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/GitProviderService.java
winklerm: Sure, np.

framework-cloud/framework-git/src/main/java/org/kie/cloud/git/gogs/GogsGitProvider.java
winklerm: Np!

sutaakar:
@jiripetrlik The PR doesn't delete projects from GitLab repo, do you have any idea what can be wrong?

sutaakar:
retest this please

PR #202, opened 2018-07-03
[7.5.x] Allow configuration of template folder URL

sutaakar:


test-cloud/pom.xml
```diff
@@ -128,8 +131,33 @@
...
+        </plugin>
       </plugins>
     </pluginManagement>
+    <resources>
+      <resource>
+        <directory>src/main/filtered-resources</directory>
+        <filtering>true</filtering>
+      </resource>
+    </resources>
+    <plugins>
```
jhrcek: Isn't maven-resources-plugin part of the build by default (so you could skip this `<plugins>` part)?

sutaakar: Not here because test-cloud module has pom packaging.

sutaakar:
PR test failures caused by incompatibility of PR builder with 7.5.x

PR #205, opened 2018-07-11
Test stabilizations after the templates switch to WebSockets

sutaakar:


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/KieServerHttpScalingIntegrationTest.java
```diff
@@ -51,6 +52,7 @@
 import org.kie.server.controller.client.KieServerControllerClient;
 
 @RunWith(Parameterized.class)
+@Ignore("Currently all templates are using WebSockets, generic scenario builder is not capable of configuring HTTP. Ignoring for now until new generic scenario builder is available.")
```
jakubschwan: This test class use also WorkbenchKieServerScenario. Right now we are ignoring all scenarios with this test class, we should ignore only scenario with Generic scenario.

sutaakar: @jakubschwan well, right now it isn't possible to configure REST connection between Kie server and controller even in templates, the WebSockets are hardcoded there. So we need to skip also classic scenarios.

PR #206, opened 2018-07-13
Fix the NPE when GitLab is not defined

sutaakar:


sutaakar:
Needs to be refactored.

sutaakar:
@jiripetrlik @jhrcek Please take a look again, implementation was adjusted. Now needed properties are verified just for a Git provider which is used, missing properties cause application to fail fast.

sutaakar:
Fixed a bug where I accidentally put GitHub credentials for GitLab, adjusted verifySystemPropertyIsSet() to catch the case when system property is empty string.

PR #209, opened 2018-07-25
Use Kjar to get GAV for container to be deployed to the Workbench

sutaakar:


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/Kjar.java
```diff
@@ -48,6 +50,10 @@ public String getVersion() {
         return version;
     }
 
+    public ReleaseId getKieServerReleaseId() {
```
jhrcek: Nice improvement. I'm just a bit confused by the name. KieServerReleaseId sounds to me like the ReleaseId of KieServer. Do we need "KieServer" in the name? If we name it just `getReleaseId` can we confuse it with something else?

sutaakar: Problem is that there are 2 different classes called ReleaseId. One is `org.kie.api.builder.ReleaseId`, used in engine, the other is `org.kie.server.api.model.ReleaseId`. I wanted to distinguish it somehow.

sutaakar: After some thoughts I decided to remove this method, it brings unnecessary dependency on Kie server API to this class.

sutaakar:
Removed getKieServerReleaseId() method to keep the Kjar enum without unneeded dependencies.

sutaakar:
Test failures not related

PR #213, opened 2018-08-09
Add smoke tests for rhpam70-prod.yaml

sutaakar:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/ClusteredWorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioImpl.java
```diff
@@ -0,0 +1,194 @@
...
+        workbenchRuntimeDeployment.scale(1);
+
+        smartRouterDeployment = createSmartRouterDeployment(project);
+        smartRouterDeployment.scale(1);
+
+        kieServerOneDeployment = createKieServerDeployment(project, "1");
+        kieServerOneDeployment.scale(1);
+
+        kieServerTwoDeployment = createKieServerDeployment(project, "2");
+        kieServerTwoDeployment.scale(1);
```
jakubschwan: Why scaling to 1? If everything is scaled to one, then this "Clustered" scenario is same as WorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenario, is it?

sutaakar: Yes, it would be basically the same.
The scaling to 1 is used to unify the state of every scenario so tests already know the scenario state and can rely on it regardless of deployed scenario.
It means that when the test starts it knows that there is just one Kie server, one Workbench... so the behaviour is same for all scenarios. It allows us to reuse same tests for many different scenarios.

sutaakar: On the second thought, is a test relevant to the scenario in case it needs to scaled first? Will think about it.

PR #214, opened 2018-08-09
Fix GIT cloning issue with GOGS

sutaakar:


framework-cloud/framework-git/src/main/java/org/kie/cloud/git/AbstractGitProvider.java
```diff
@@ -41,7 +43,11 @@ protected String generateRepositoryName(String repositoryPrefixName) {
     protected void pushToGitRepository(String httpUrl, String repositoryPath,
             String username, String password) {
         try {
-            Git git = Git.init().setDirectory(new File(repositoryPath)).call();
+            File originalDirectory = new File(repositoryPath);
+            File tempDirectory = Files.createTempDirectory("openshift-git").toFile();
+            FileUtils.copyDirectory(originalDirectory, tempDirectory);
```
jhrcek: @sutaakar Can you please explain why we need this?

sutaakar: @jhrcek Currently the projects are placed in target/test-classes/kjar-sources folder. When a test wants to push the project into Gogs repo, it has to create a Git repo from the folder which it wants to push and then push it to remote repo.
Tests are using 2 ways to create and push repo. One approach is to create repo from whole kjar-sources folder, push it and then test method use some specific folder in remote repo. Other approach is to push some folder from kjar-sources. The result is that over time target/test-classes/kjar-sources contains initialized repo and some of its folders also contain initialized repo.
This causes issues with Gogs as we cannot access subfolder of kjar-sources in case it also contains initialized repo. 

The solution is to copy required repo folder to temporary location to be initialized there so we can avoid hierarchical git repos.

jhrcek: @sutaakar thanks for the explanation. It would be nice to have this kind of explanation in the commit message. The code change looks innocent, but the reasons are not trivial, so it might be good to put a sentence or two into message body..

sutaakar: @jhrcek commit message updated

PR #215, opened 2018-08-10
Move HTTPS tests to scenario smoke tests

sutaakar:
SSO tests are moved to separate test classes.
Main intention is to speed up test suite execution by running multiple tests against single deployed scenario.


sutaakar:
Test failures not related

PR #226, opened 2018-09-10
Temporary use SSO image stream pointing to old registry

sutaakar:


sutaakar:
Test failures not related (Timer and Image version tests)

PR #233, opened 2018-09-20
Centralize product-specific logic into ProjectProfile

jhrcek:
Problems I'm trying to solve with this PR:
- remove redundant `active.test.profile` which has the same information as `template.project`
- reduce the proliferation of `if product1 else if product2 ...`

Changes made:
- Removed `TemplateSelector` as it's no longer useful (after DatabaseProfile was removed)
- Pulled out `TemplateSelector.Project` into separate top level Enum `ProjectProfile`.
- Removed Jakub's `ActiveTestProfile` and replaced all usage with `ProjectProfile`  which contains the same information (drools vs. jbpm)
- Replaced repeated all `if drools .. else if jbpm ..` with `ProductSpecificPropertyNames` (centralize decision about product in single place)
- Made all product-specific property names package private, forcing clients to use `ProductSpecificPropertyNames` to prevent errors [like these](https://github.com/kiegroup/kie-cloud-tests/blob/master/framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/settings/builder/KieServerMySqlSettingsBuilderImpl.java#L142-L146) and reduce some boilerplate.


framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/builder/WorkbenchKieServerPersistentScenarioBuilderImpl.java
```diff
@@ -38,16 +40,12 @@ public WorkbenchKieServerPersistentScenarioBuilderImpl() {
...
         envVariables.put(OpenShiftTemplateConstants.MAVEN_REPO_PASSWORD, DeploymentConstants.getWorkbenchPassword());
-        if (ActiveTestProfile.isJbpm()) {
-            envVariables.put(OpenShiftTemplateConstants.BUSINESS_CENTRAL_HTTPS_SECRET, OpenShiftConstants.getKieApplicationSecretName());
-        } else if (ActiveTestProfile.isDrools()) {
-            envVariables.put(OpenShiftTemplateConstants.DECISION_CENTRAL_HTTPS_SECRET, OpenShiftConstants.getKieApplicationSecretName());
-        }
         envVariables.put(OpenShiftTemplateConstants.KIE_SERVER_HTTPS_SECRET, OpenShiftConstants.getKieApplicationSecretName());
+        envVariables.put(propertyNames.workbenchMavenUserName(), DeploymentConstants.getWorkbenchMavenUser());
+        envVariables.put(propertyNames.workbenchMavenPassword(), DeploymentConstants.getWorkbenchMavenPassword());
+        envVariables.put(propertyNames.workbenchMavenPassword(), OpenShiftConstants.getKieApplicationSecretName());
```
sutaakar: Should be propertyNames.workbenchHttpsSecret() here.

jhrcek: Good catch. Fixing..

jhrcek:
Added license headers in new files.

PR #236, opened 2018-09-25
Operator adjustments and implementation

sutaakar:
This is just initial implementation as Kie Operator supports just one scenario with hardcoded properties.

### DIFF
--- a/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/resources/OpenShiftResourcePropertiesLoader.java
+++ b/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/resources/OpenShiftResourcePropertiesLoader.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.kie.cloud.openshift.operator.resources;
+
+import java.util.Properties;
+
+import org.kie.cloud.openshift.util.PropertyLoader;
+
+class OpenShiftResourcePropertiesLoader {
+
+    private OpenShiftResourcePropertiesLoader() {
+        // Util class
+    }
+
+    private static final Properties resourceUrlProperties = new Properties();
+
+    static {
+        loadResourcePropertiesFromResources();
+    }
+
+    static Properties getProperties() {
+        return resourceUrlProperties;
+    }
+
+    private static void loadResourcePropertiesFromResources() {
+        resourceUrlProperties.putAll(PropertyLoader.loadProperties(OpenShiftResourcePropertiesLoader.class, "resources.properties"));
+    }
+}
--- a/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/WorkbenchKieServerScenarioBuilderImpl.java
+++ b/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/WorkbenchKieServerScenarioBuilderImpl.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
+ *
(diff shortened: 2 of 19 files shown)


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/resources/OpenShiftResourcePropertiesLoader.java
```diff
@@ -0,0 +1,41 @@
...
+    static {
+        loadResourcePropertiesFromResources();
+    }
+
+    static Properties getProperties() {
+        return resourceUrlProperties;
+    }
+
+    private static void loadResourcePropertiesFromResources() {
+        resourceUrlProperties.putAll(PropertyLoader.loadProperties("/resources.properties"));
```
jhrcek: I'd recommend putting these properties to the package corresponding to this class
(src/main/resources/org/kie/cloud/openshift/operator/resources) and using relative path (`loadProperties("resources.properties")`) because it's not clear from the filename what kind of resources are there. But leaving this up to you.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/PropertyLoader.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Process properties and replace any variable with appropriate value. The value can be retrieved from another property in properties.
+     *
+     * @param properties Properties to be processed.
+     * @return Processed properties.
+     */
+    private static Properties filterProperties(Properties properties) {
+        Pattern replaceTagPattern = Pattern.compile(".*\\$\\{(.*)\\}.*");
```
jhrcek: Perpahs compile just once and place the pattern in a static field?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/PropertyLoader.java
```diff
@@ -0,0 +1,84 @@
...
+        Properties processedProperties = new Properties();
+
+        properties.forEach((k,v) -> {
+            String value = (String) v;
+            Matcher matcher = replaceTagPattern.matcher(value);
+            if(matcher.matches()) {
+                // If property tag is provided using system properties then it is replaced automatically when loading properties by Java.
+                // This step is used just in case the replacement is another property from loaded properties.
+                String propertyValue = properties.getProperty(matcher.group(1));
+                processedProperties.put(k, value.replace("${" + matcher.group(1) + "}", propertyValue));
```
jhrcek: I hope the properties won't have several levels of indirection.
e.g. 
A=${B}
B=${C}

If we don't want to allow that I'd recommend asserting that `propertyValue` doesn't contain any match of `replaceTagPattern` and if yes, throw an exception (otherwise you'd have to do the regex replacement recursively, but then you'd have to deal with potentially cyclic dependencies like
a=${b}
b=${a}
etc..)

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/resources/OpenShiftResourcePropertiesLoader.java
sutaakar: Functionality refactored, now it is possible to use relative path. Property files moved to locations related to loader classes.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/PropertyLoader.java
sutaakar: Good tip, thanks. Implemented

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/PropertyLoader.java
sutaakar: Added validation. There is just one level of indirection allowed.

sutaakar:
retest this please

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/WorkbenchKieServerScenarioBuilderImpl.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public WorkbenchKieServerScenario build() {
+        return new WorkbenchKieServerScenarioImpl();
+    }
+
+    @Override
+    public WorkbenchKieServerScenarioBuilder withExternalMavenRepo(String repoUrl, String repoUserName, String repoPassword) {
+        // TODO: currently skipped to make tests running somehow, needs to be implemented
```
jhrcek: (optional) If throwing UnsupportedOperationException is not acceptable, maybe at least log this as a warning so it's a bit harder to forget?

sutaakar:
Test failures not related.

PR #237, opened 2018-09-26
Log repository creation

sutaakar:
Used to narrow down what projects are created when the HTTP 500 is returned


sutaakar:
Test failure not related

PR #250, opened 2018-10-19
Add support for CORS headers to WorkbenchKieServerScenario

sutaakar:
That is the only scenario which has CORS template properties.


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/WorkbenchKieServerScenarioBuilderImpl.java
```diff
@@ -44,4 +46,29 @@ public WorkbenchKieServerScenarioBuilder withExternalMavenRepo(String repoUrl, S
     public WorkbenchKieServerScenarioBuilder withKieServerId(String kieServerId) {
         throw new UnsupportedOperationException("Not implemented yet");
     }
+
+    @Override
+    public WorkbenchKieServerScenarioBuilder withAccessControlAllowCredentials(boolean allowCredentials) {
+        throw new UnsupportedOperationException("Not implemented yet");
```
jhrcek: Will this ever be implemented? If no, you should probably change the exception message.
Also having to use `UnsupportedOperationException` is usually a sign that the interface that this builder implements is "too broad" and should probably be split into smaller sub-interfaces.

PR #254, opened 2018-10-30
RHPAM-1643: Remove rhpam72-sit.yaml template from framework and tests

sutaakar:
@jhrcek @jiripetrlik FYI the rhpam72-sit.yaml template was deleted on master as it is almost equal to rhpam72-prod.yaml.


sutaakar:
The remaining failures are related to version mismatch (expected) and replica adjustments made in https://github.com/jboss-container-images/rhpam-7-openshift-image/pull/168 .
The only way I see how it could be fixed is to create a template using oc client, to use it so the XTF version needs to he raised to 0.9.

PR #257, opened 2018-10-31
Process and create template using oc client

sutaakar:
Workaround for fabric8 client limitation - cannot handle template placeholders for fields expecting numeric value.
Also deleted processTemplateAndCreateResources which shouldn't be used anywhere.

It shouldn't have any side effect.


sutaakar:
Test failure not related

PR #267, opened 2018-12-17
Remove version adjustments in parent pom

sutaakar:
As versions were adjusted in Kie parent pom.
Latest changes on master branch broke down compilation.


sutaakar:
Test failures not related.

PR #273, opened 2019-01-15
Tests adjustments for APB framework 

jakubschwan:
Commits in PR will be squashed with merge of this PR


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/resource/impl/ProjectImpl.java
```diff
@@ -100,34 +101,43 @@ public void processTemplateAndCreateResources(URL templateUrl, Map<String, Strin
         }
     }
 
+    static Semaphore semaphore = new Semaphore(1);
```
sutaakar: The synchronized on ProjectImpl methods wasn't enough it this case?
What is the source of concurrency here?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/ApbImageGetter.java
```diff
@@ -29,6 +29,8 @@
     static {
         OpenShiftConfig openShiftConfig = new OpenShiftConfigBuilder()
                 .withDisableApiGroupCheck(true)
+                .withUsername("admin") //TODO: insert it as system property in pom.xml // OpenShiftConstants.getOpenShiftAdminUserName()
```
sutaakar: Why are the OpenShiftConstants commented out?

test-cloud/test-cloud-remote/pom.xml
```diff
@@ -125,6 +130,84 @@
...
+                  </configuration>
+                </execution>
+                <execution>
+                  <id>importJavaKeystore</id>
+                  <goals>
+                    <goal>importKeystore</goal>
+                  </goals>
+                  <phase>package</phase>
+                  <configuration>
+                    <srckeystore>${java.home}/lib/security/cacerts</srckeystore>
```
sutaakar: What certificates are expected to be in this keystore?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/AbstractMethodIsolatedCloudIntegrationTest.java
```diff
@@ -32,7 +33,11 @@
 
     @Before
     public void initializeDeployment() {
-        deploymentScenario = createDeploymentScenario(deploymentScenarioFactory);
+        try {
+            deploymentScenario = createDeploymentScenario(deploymentScenarioFactory);
+        } catch (UnsupportedOperationException ex) {
+            Assume.assumeNoException(ex);
```
sutaakar: Maybe it would be better to apply category, something like Apb class, to differentiate tests which can be executed with APB and which not. Handling exceptions in initialization code makes the test less readable and understandable.

test-cloud/test-cloud-remote/pom.xml
jakubschwan: In this keystore are certificates for java. Newly created certificate in previous execution is not able to communicate with all system.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/AbstractMethodIsolatedCloudIntegrationTest.java
jakubschwan: +1 for the category. This assume also do not work correctly, I found out that if assume is used in parent class or called from parent class, tests are not skipped.

PR #275, opened 2019-01-17
Use resource filtering for images and templates version and branch

rsynek:
If it's ok, I will cherry-pick into 7.14.x


rsynek:
@sutaakar the check fails as there are no such templates on master branch - which was the case even before opening this PR. Do I understand it correctly that this check currently cannot work for master branch?

PR #290, opened 2019-02-18
Raise prefix version to 73

sutaakar:


sutaakar:
Skipping PR build as it uses upstream image streams (not available for latest images).

PR #297, opened 2019-03-01
Add support for pre deployment listeners

sutaakar:


rsynek:
@sutaakar could you please link a Jira or put there some description what is this enhancement about and what was the motivation for it? Thanks!

sutaakar:
@rsynek You can find the details and motivation in https://issues.jboss.org/browse/BAQE-890

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/DeploymentScenario.java
```diff
@@ -15,11 +15,13 @@
 package org.kie.cloud.api.scenario;
 
 import java.util.List;
+import java.util.Optional;
```
jakubschwan: Unused import

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/builder/KieScenarioBuilderImpl.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
jakubschwan: Wrong year in license header, please fix it in all new files.

rsynek:
Overall looks good, just two concerns:
- more than just pre-deploy event may be needed (e.g. undeploy?), so the `Consumer` probably won't do the job
- the `OpenShiftScenario` calls the listeners in the `deploy` method, while each scenario is required to override the method and call `super.deploy()` as the very first thing to do. This is a very fragile contract. If you need to ensure all scenarios follow a certain process, please consider making the deploy/undeploy methods final and offer a way to hook their code in without the risk of overriding the process.

sutaakar:
@rsynek @jakubschwan Comments addressed, please take a look again.

PR #305, opened 2019-03-08
Add test coverage for managed template

sutaakar:
Had to adjust test providers to be able to run tests without creating of containers (to avoid respins).

PR #306, opened 2019-03-08
[7.18.x] Add test coverage for managed template

sutaakar:
Cherrypick of https://github.com/kiegroup/kie-cloud-tests/pull/305


jakubschwan:
@sutaakar just add same changes as into #305

sutaakar:
Updated

PR #312, opened 2019-03-13
Stabilization of ClusteredWorkbenchRuntimeClusteredKieServerDatabaseS…

sutaakar:
…cenario


sutaakar:
@jakubschwan Updated, not the scenario keeps 3 pods for both Kie server and Monitoring console

PR #327, opened 2019-03-27
Kie Operator alignment for latest changes

sutaakar:


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/ClusteredWorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioImpl.java
```diff
@@ -59,49 +57,35 @@ protected void deployCustomResource() {
...
+        kieServerDeployment = createKieServerDeployment(project);
+        databaseDeployment = createDatabaseDeployment(project);
+
+        logger.info("Waiting until all services are created.");
         try {
-            new SimpleWaiter(() -> project.getOpenShiftUtil().getServices().size() == 9).timeout(TimeUnit.MINUTES, 1).execute();
+            new SimpleWaiter(() -> workbenchRuntimeDeployment.isReady()).timeout(TimeUnit.MINUTES, 1).execute();
+            new SimpleWaiter(() -> smartRouterDeployment.isReady()).timeout(TimeUnit.MINUTES, 1).execute();
+            new SimpleWaiter(() -> kieServerDeployment.isReady()).timeout(TimeUnit.MINUTES, 1).execute();
+            new SimpleWaiter(() -> databaseDeployment.isReady()).timeout(TimeUnit.MINUTES, 1).execute();
```
jakubschwan: Shouldn't we check if DB is ready before the Kie Server deployment? E.g. in case when external DB used.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/ClusteredWorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioImpl.java
```diff
@@ -118,22 +102,22 @@ public SmartRouterDeployment getSmartRouterDeployment() {
...
     @Override
     public KieServerDeployment getKieServerOneDeployment() {
-        return kieServerOneDeployment;
+        return kieServerDeployment;
     }
 
     @Override
     public KieServerDeployment getKieServerTwoDeployment() {
-        return kieServerTwoDeployment;
+        return kieServerDeployment;
```
jakubschwan: If there is only one Kie Server and Dabatase it should be good to add some logger message with warning when is get second deployment.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
```diff
@@ -60,33 +58,34 @@ private void deployOperator() {
...
             logger.info("Creating custom resource definitions from " + OpenShiftResource.CRD.getResourceUrl().toString());
-            OpenShiftBinaryClient.getInstance().executeCommand("CRD failed.", "create", "-f", OpenShiftResource.CRD.getResourceUrl().toString());
+            OpenShiftBinaryClient instance = OpenShiftBinaryClient.getInstance();
+            try (ProcessExecutor executor = new ProcessExecutor()) {
+                executor.executeProcessCommand(instance.getOcBinaryPath().toString() + " --config=" + instance.getOcConfigPath().toString() + " create -n " + getNamespace() + " -f " + OpenShiftResource.CRD.getResourceUrl().toString());
+            }
         }
     }
 
+    private void createServiceAccountInProject(Project project) {
```
jakubschwan: Similar operations are in APB. It would be good to move it in framework parent and share this functionality in both places.
For now we can leave it here and open a new PR with changes for creating Service acc and role.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/ClusteredWorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioImpl.java
sutaakar: This check makes sure that the service is created, there is no need to keep an order.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
sutaakar: I wouldn't share it explicitly as the Operator installation can change. Current approach is using direct installation into the project, but in the future we may switch to installation using OLM.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/ClusteredWorkbenchRuntimeSmartRouterTwoKieServersTwoDatabasesScenarioImpl.java
sutaakar: Added warning. Please take a look

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
tchughesiv: @sutaakar agreed on OLM and, honestly, that should probably be the main (only) install method tested here... since it will be the only supported method for users.

PR #359, opened 2019-04-25
Some exclusions are needed to avoid duplicated classes issue

gmunozfe:
There are duplicated classes that have to be excluded for full maven install execution


sutaakar:
I stopped the PR job as it is not configured properly.

sutaakar:
@gmunozfe Please forward port this PR also to master branch.

PR #366, opened 2019-05-06
[7.18.x] Add exclusions to avoid duplicate tests classes

jakubschwan:
Excluded resteasy-jackson-provider instead of resteasy-jackson2-provider
because of failing RH SSO integration tests.

### DIFF
--- a/framework-cloud/framework-cloud-common/pom.xml
+++ b/framework-cloud/framework-cloud-common/pom.xml
@@ -49,6 +49,10 @@
           <groupId>javax.xml.stream</groupId>
           <artifactId>stax-api</artifactId>
         </exclusion>
+        <exclusion>
+            <groupId>org.jboss.resteasy</groupId>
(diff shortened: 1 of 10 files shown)


framework-cloud/framework-cloud-common/pom.xml
```diff
@@ -87,6 +97,7 @@
     <dependency>
       <groupId>cz.xtf</groupId>
       <artifactId>utilities</artifactId>
+      <!-- -->
```
gmunozfe: You may delete this line

framework-cloud/framework-cloud-common/pom.xml
```diff
@@ -96,10 +107,12 @@
             <groupId>javax.jms</groupId>
             <artifactId>jms-api</artifactId>
           </exclusion>
+          <!--
```
gmunozfe: You may delete all the lines inside the comment for clarity

PR #374, opened 2019-05-08
New test-cloud-performance module

gmunozfe:
@sutaakar @MarianMacik  first scenario for EJB Timers in Openshift


gmunozfe:
thanks @sutaakar  you're right, just removed that submodule in the bottom.

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
```diff
@@ -0,0 +1,358 @@
...
+           int sizeStarted = queryOldValue(i, pod);
+           startedHostNameDistribution.put(pod, startedHostNameDistribution.get(pod)+sizeStarted);
+          }
+        }
+        
+        logger.info("Processes were completed with this distribution: {}", completedHostNameDistribution);
+        logger.info("Processes were started with this distribution: {}", startedHostNameDistribution);       
+    }
+    
+    private void scaleKieServerTo(int count) {
```
MarianMacik: I am afraid that this is only called if there is more than one pod to be used in scenario. If only 1 pod is used, then the pod's name won't be added to `pods` collection so statistics gathering won't work?

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
```diff
@@ -0,0 +1,358 @@
...
+        gatherAndAssertStatistics();
+        writeCSV();
+        logger.info("============================= STATISTICS GATHERED =============================");       
+    }
+
+    private void writeCSV() throws IOException {
+        List<String> headerCompleted = new ArrayList<String>();
+        for (int i=1; i<=pods.size(); i++) {
+            headerCompleted.add("Completed Pod"+i);
+        }
```
MarianMacik: Is this used anywhere? It was probably meant to be used as a header?

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
```diff
@@ -0,0 +1,358 @@
...
+       
+        Map<String, Integer> startedHostNameDistribution = new HashMap<>();
+        completedHostNameDistribution = new HashMap<>();
+       
+        for (String pod : pods) {
+            completedHostNameDistribution.put(pod, 0);
+            startedHostNameDistribution.put(pod, 0);
+        }
+        
+        for (int i = 0; i < numberOfPages-1; i++) {
```
MarianMacik: I assume this is because multiplications of 5000 will query here the non-existent page, which is added because we want to first check if there is really for example 10k processes and not more.

However, this -1 will not work for example for 14 000 processes. 14 000/5000 will return 2, and +1 is 3 on line 300. So here you will only go through pages 0 and 1. I suggest that for first check if there are as many processes completed as we want, we should use for example:
```
int numberOfPages = ((Double)Math.ceil(PROCESSES_COUNT / 5000)).intValue(); // number of pages for normal query
```
And then in the first check we just add +1 to query one more page.

Or we can leave it as it is now and don't decrease numberOfPages by 1 on line 326, because it should have no effect if we want page which does not exist - it should give us 0 results, in fact, we do it like that in the first check (line 304-3007) and it worked as far as I know.

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
```diff
@@ -0,0 +1,358 @@
...
+          for (String pod : pods) {
+           int sizeCompleted = queryServicesClient.findProcessInstancesByVariableAndValue("hostName", pod, COMPLETED_STATUS, i, 5000).size();
+           completedHostNameDistribution.put(pod, completedHostNameDistribution.get(pod)+sizeCompleted);
+           
+           int sizeStarted = queryOldValue(i, pod);
+           startedHostNameDistribution.put(pod, startedHostNameDistribution.get(pod)+sizeStarted);
```
MarianMacik: Please also use correct formatting in this class in situations like `startedHostNameDistribution.get(pod)+sizeStarted` etc.

test-cloud/test-cloud-common/src/main/java/org/kie/cloud/tests/common/client/util/RunnableWrapper.java
```diff
@@ -0,0 +1,61 @@
+package org.kie.cloud.tests.common.client.util;
+
+import java.time.Duration;
+import java.time.Instant;
+
+/**
+ * Wrapper of Runnable we want to run for a fixed period of time
+ * or a fixed number of iterations, whan
```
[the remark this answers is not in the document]
gmunozfe: Fixed

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
```diff
@@ -0,0 +1,358 @@
...
+        
+        logger.info("Starting timers-testing.OneTimerDate");
+        
+        startAndWaitForStartingThreads(STARTING_THREADS_COUNT, maxDuration, PROCESSES_PER_THREAD, getStartingRunnable(CONTAINER_ID, "timers-testing.OneTimerDate", params));
+        startingTime = Duration.between(startTime, Instant.now()).toString();
+        logger.info("Starting processes took: {}", startingTime);
+
+        assertThat(Instant.now()).isBefore(fireAtTime.toInstant());
+
+        Duration waitForCompletionDuration = Duration.between(Instant.now(), fireAtTime).plus(Duration.of(25/*15 40*/, ChronoUnit.MINUTES));
```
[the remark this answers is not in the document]
gmunozfe: Right, left the most restricted one, 40

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
```diff
@@ -0,0 +1,358 @@
...
+        logger.info("Waiting for process instances to be completed, max waiting time is {}", waitForCompletionDuration);
+        waitForAllProcessesToComplete(waitForCompletionDuration);
+        processTime = Duration.between(fireAtTime.toInstant(), Instant.now()).toString();
+        logger.info("Process instances completed, took approximately {}", processTime);
+    }
+    
+    private OffsetDateTime calculateFireAtTime() {
+        OffsetDateTime currentTime = OffsetDateTime.now();
+        long offsetInSeconds = Math.round(PROCESSES_COUNT / STARTING_THREADS_COUNT / PERF_INDEX);
+        offsetInSeconds = offsetInSeconds<45 ? 45 : offsetInSeconds;
```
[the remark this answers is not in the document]
gmunozfe: Could be useful for testing few processes, but it can be removed also, wdyt?

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
gmunozfe: right, should be the same for all cases: I've added this case with one pod also to `pods` collection

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
gmunozfe: Removed the -1 on line 326 as you suggests as second option

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
gmunozfe: not sure what you refer to in this case. Could you give me a little more detail?

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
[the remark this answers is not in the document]
MarianMacik: OK, can be left, but please extract it to a constant, something like minimum offset or so...

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
MarianMacik: I mean there should be white spaces between operator and operands.

jakubschwan:
Looks good. I have one idea how to enhance the results. Right now is set only memory limit for the Kie Server in template by the env. variable KIE_SERVER_MEMORY_LIMIT. It would be good to set also memory requests and CPU limit with request. By this value we can specify how many resources is used for one container/pod. Right now this values are not part of the template, but we can add them to the Deployment config. With updated DC are pods automatically restarted. More about resources is in OS doc [Quotas and Limit Ranges](https://docs.openshift.com/container-platform/3.11/dev_guide/compute_resources.html)

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/settings/builder/KieServerHttpsS2ISettingsBuilderImpl.java
```diff
@@ -163,4 +164,17 @@ public KieServerS2ISettingsBuilder withKieServerSecret(String secret) {
...
+    @Override
+    public KieServerS2ISettingsBuilder withTimerServiceDataStoreRefreshInterval(
+            Duration timerServiceDataStoreRefreshInterval) {
+        envVariables.put(OpenShiftTemplateConstants.TIMER_SERVICE_DATA_STORE_REFRESH_INTERVAL, Long.toString(timerServiceDataStoreRefreshInterval.toMillis()));
+        return this;
+    }
+    
+    @Override
+    public KieServerS2ISettingsBuilder withKieServerMemoryLimit(String limit) {
+        envVariables.put("KIE_SERVER_MEMORY_LIMIT", limit);
```
jakubschwan: please add key KIE_SERVER_MEMORY_LIMIT to the OpenShiftTemplateConstants class.

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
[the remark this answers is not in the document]
gmunozfe: thanks for the hint, done

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
gmunozfe: No, removed as this was for an attempt to add headers (but the Apache CSV library doesn't allow to have a variable number of headers, as this case)

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
gmunozfe: Thanks for the clarification. Done

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
```diff
@@ -169,6 +169,20 @@ public void resetRouterTimeout() {
             .endMetadata()
             .done();
     }
+    
+    @Override
+    public void setRouterBalance(String balance) {
+        // Route has a same name as its service.
+        String routeName = getServiceName();
```
jakubschwan: This is true only for HTTP route, HTTPS route has prefix "secure-" in name. If you want to set Router balance. It will be better to search for routes by label service [in template configured here](https://github.com/jboss-container-images/rhpam-7-openshift-image/blob/master/templates/rhpam74-prod-immutable-kieserver.yaml#L516). By this you should be able to find all routes for the service and then you can set to all of them router balance.

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/settings/builder/KieServerHttpsS2ISettingsBuilderImpl.java
gmunozfe: Ok, thanks, done

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
MarianMacik: Well, I think that this is only about the number of columns and this is for one run of the tests fixed, so it shouldn't be an issue to add headers, should it? But we can improve that in a separate PR if it is for a longer debate.

gmunozfe: Yes, we can leave it as a future enhancement. I was trying (without success) to add a variable number of headers but it seems like [Apache library](http://commons.apache.org/proper/commons-csv/user-guide.html#Printing_with_headers) is not prepared to.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
gmunozfe: Fixed also for setRouterBalance/resetRouterBalance

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
```diff
@@ -205,6 +223,13 @@ private Instance createInstance(Pod pod) {
     protected Optional<URI> getWebSocketRouteUri(String serviceName) {
         return getRouteUri(Protocol.ws, serviceName);
     }
+    
+    protected RouteList getRoutes(){
+        return util.client()
+                   .routes()
+                   .withLabel("key", "value")
```
jakubschwan: key should be "service" and value should be getServiceName()

gmunozfe: sure.... fixed

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
```diff
@@ -0,0 +1,368 @@
...
+        deploymentScenario.getKieServerDeployments().get(0).setRouterTimeout(Duration.ofMinutes(ROUTER_TIMEOUT));
+        deploymentScenario.getKieServerDeployments().get(0).setRouterBalance(ROUTER_BALANCE);
+        
+        Map<String, String> requests = new HashMap<String, String>();
+        requests.put(CPU, System.getProperty("requests.cpu","1000m"));
+        requests.put(MEMORY, System.getProperty("requests.memory","1Gi"));
+        Map<String, String> limits = new HashMap<String, String>();
+        limits.put(CPU, System.getProperty("limits.cpu","4000m"));
+        limits.put(MEMORY, System.getProperty("limits.memory","4Gi"));
+        deploymentScenario.getKieServerDeployments().get(0).setResources(requests, limits);
```
jakubschwan: to be sure change is done and pod is ready, it would be also good to add there waiter for Kie Server to be ready. something like: deploymentScenario.getKieServerDeployments().get(0).waitForScale()

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
```diff
@@ -144,30 +147,68 @@ protected void waitUntilAllPodsAreReadyAndRunning(int expectedPods) {
...
+        }
+    }
+    
+    @Override
+    public void setResources(Map<String,String> requests, Map<String,String> limits) {
         util.client()
-            .routes()
-            .withName(routeName)
+            .deploymentConfigs()
+            .withName(getServiceName())
```
jakubschwan: this can be confusing to read. Can you put there a comment
```suggestion
            .withName(getServiceName()) // Deployment config has a same name as its service.
```

test-cloud/test-cloud-performance/src/test/java/org/kie/cloud/integrationtests/s2i/KieServerS2iJbpmEJBTimerPerfIntegrationTest.java
gmunozfe: ok, I wasn't aware of changes of the DC implies that pods are also restarted, so we need to wait for this to be ready
Done, and added also a comment to reflect this. Thx!

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
gmunozfe: Done

PR #375, opened 2019-05-10
KIECLOUD-184: Adjust integration tests to check respins

sutaakar:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/KieServerDeploymentImpl.java
```diff
@@ -90,4 +99,44 @@ public String getServiceName() {
             RouterUtil.waitForRouter(getUrl());
         }
     }
+
+    @Override
+    public void waitForContainerRespin() {
+        // Expecting the config map name to be same as service name. Needs to be adjusted in case this changes!!!!
+        ConfigMap kieServerDeployments = getOpenShift().getConfigMap(getServiceName());
```
jakubschwan: Isn't there a label service name in the config map? We could find the config map by this label then.

sutaakar:
Adjusted implementation according to comment. Now the config map is retrieved based on owner reference which is equal to deployment config name.

PR #394, opened 2019-06-14
Tests for leader election in HA CEP

jiripetrlik:


framework-cloud/framework-openshift-strimzi/src/main/java/org/kie/cloud/strimzi/deployment/KafkaDeployment.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public KafkaDeployment(String clusterName, Project project) {
+        super(project);
+
+        this.clusterName = clusterName;
+    }
+
+    @Override
+    public void waitForScale() {
+        SupplierWaiter statefulSetExistWaitter = new SupplierWaiter(
```
sutaakar: Waitter -> Waiter

test-cloud/test-cloud-ha-cep/src/test/java/org/kie/cloud/hacep/HACEPLeaderElectionIntegrationTest.java
```diff
@@ -0,0 +1,86 @@
...
+                    .getDeployments()
+                    .get(0)
+                    .getInstances()
+                    .stream().map(instance -> instance.getName())
+                    .collect(Collectors.toList());
+
+            final ConfigMap leadersConfigMap = project.getOpenShift().configMaps().withName(LEADERS_CONFIG_MAP).get();
+            Assertions.assertThat(leadersConfigMap.getData()).containsKey(LEADER_POD_KEY);
+            Assertions.assertThat(leadersConfigMap.getData().get(LEADER_POD_KEY)).isIn(podNames);
+        } catch (Exception e) {
```
sutaakar: You don't have to catch exception here, just throw it in test method. It will be caught by JUnit.

jiripetrlik:
@sutaakar 
Thank you for review. All comments should be fixed now.

framework-cloud/framework-openshift-strimzi/src/main/java/org/kie/cloud/strimzi/resources/KafkaClusterBuilder.java
```diff
@@ -0,0 +1,111 @@
...
+        this.name = name;
+    }
+
+    public KafkaClusterBuilder withKafkaReplicas(final int kafkaReplicas) {
+        this.kafkaReplicas = kafkaReplicas;
+
+        return this;
+    }
+
+    public KafkaClusterBuilder addKafkaConfigItem(final String key, final String value) {
```
jakubschwan: can be also handy to have method 'addKafkaConfigItems(Map<String, String>)' to add all items from map. Later in code are set same properties 3 times. To avoid configuration issue it would be better to add these same values to one map and then set it to all 3 topics.

PR #464, opened 2019-08-29
[7.26.x] Introduce deployment config name for deployments

sutaakar:
Old code expected deployment config name to be same as service name.
That is not true any more, for example for AMQ.

### DIFF
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
@@ -88,9 +88,13 @@ public abstract class OpenShiftDeployment implements Deployment {
 
     public abstract String getServiceName();
 
+    public String getDeploymentConfigName() {
+        return getServiceName();
+    }
+
     @Override
     public void scale(int instances) {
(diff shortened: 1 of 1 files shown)


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
```diff
@@ -88,9 +88,13 @@ public void deleteInstances(List<Instance> instances) {
 
     public abstract String getServiceName();
 
+    public String getDeploymentConfigName() {
```
jakubschwan: Shouldn't be better to have this method abstract too, so for every Deployment will be this set as needed? As you mentioned in comment of this PR, we need this for AMQ and in this PR I don't see change form AMQ deployment impl.

sutaakar: The AMQ usage will come in next PR. I try to keep PRs as small as possible to amke reviewes easier.
The reason of not making this method abstract is that for most deployments their deployment config name is equal to service name. Making this method abstract would force all deployments to implement this method.This approach allows deployment to override method only if they need to define custom deployment config name.

PR #482, opened 2019-09-09
[7.26.x] Implement Operator support for ImmutableKieServerScenarioBuilder

sutaakar:
Also fixing errors and changes related to OpenShift 4.2


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/ImmutableKieServerScenarioImpl.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    @Override
+    public KieServerDeployment getKieServerDeployment() {
+        return kieServerDeployment;
+    }
+
+    @Override
+    public List<Deployment> getDeployments() {
+        List<Deployment> deployments = new ArrayList<>(Arrays.asList(kieServerDeployment));
```
jakubschwan: Add missing SSO Deployment
```suggestion
        List<Deployment> deployments = new ArrayList<>(Arrays.asList(kieServerDeployment,ssoDeployment));
```

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/ImmutableKieServerScenarioBuilderImpl.java
```diff
@@ -0,0 +1,206 @@
...
+        for (Server server : kieApp.getSpec().getObjects().getServers()) {
+            server.addEnv(new Env(ImageEnvVariables.KIE_SERVER_ID, kieServerId));
+        }
+        return this;
+    }
+
+    @Override
+    public ImmutableKieServerScenarioBuilder withHttpKieServerHostname(String hostname) {
+        for (Server server : kieApp.getSpec().getObjects().getServers()) {
+            server.addEnv(new Env(ImageEnvVariables.HOSTNAME_HTTP, hostname));
```
jakubschwan: Shouldn't we add there a suffix to make a different hostnames for more servers? (In similar way how are resolved sso clients). Or is that add in operator, or it can be used only when is set only one Kie Server deployment?

sutaakar: Well, in this scenario there is just one Kie server deployment available. I kept there for loop as kind of safety measure (or just a copy paste?).

PR #487, opened 2019-09-13
[7.26.x] Implement AMQ support for operator RHDM scenario

sutaakar:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
```diff
@@ -92,6 +94,18 @@ public String getDeploymentConfigName() {
         return getServiceName();
     }
 
+    protected String getDeploymentConfigName(OpenShift openShift, Pattern regexp) {
+        // Try to find deployment config name from all available deployment configs
+        List<DeploymentConfig> deploymentConfigs = openShift.getDeploymentConfigs();
+        for (DeploymentConfig deploymentConfig : deploymentConfigs) {
+            if (regexp.matcher(deploymentConfig.getMetadata().getName()).matches()) {
```
jakubschwan: What about case when is more then one suitable match found? I think it would be better to replace this with stream witch filter out matches and if there are more then one we should throw some exception.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/s2i/jms/KieServerS2iAmqOptaplannerIntegrationTest.java
```diff
@@ -56,7 +55,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 
 @RunWith(Parameterized.class)
-@Category({ApbNotSupported.class, OperatorNotSupported.class})
+@Category({ApbNotSupported.class})
```
jakubschwan: I think there are 2 more test classes from where should be removed OperatorNotSupported category.

jakubschwan: Even that PR is related to RHDM only, test of RHPAM should be skipped as there is JBPMOnly category.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java
sutaakar: Implemented

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/s2i/jms/KieServerS2iAmqOptaplannerIntegrationTest.java
sutaakar: jBPM tests are skipped automatically through UnsupportedOperationException check

PR #504, opened 2019-10-10
Router adjustments

sutaakar:


framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/RouterUtil.java
```diff
@@ -41,40 +34,22 @@
...
-                HttpClient client = HttpClientBuilder.create().build();
-                HttpResponse response = client.execute(request);
-
-                if (response.getStatusLine().getStatusCode() != ROUTER_CODE) {
-                    return;
-                }
-
-                String responseContent = IOUtils.toString(response.getEntity().getContent(), "UTF-8");
-                if (!responseContent.contains(ROUTER_MESSAGE)) {
+                if (Https.getCode(urlString) != ROUTER_CODE) {
```
jakubschwan: I think it would be good what type of code we check. Change the name of is enough. e.g.:
```suggestion
                if (Https.getCode(urlString) != ROUTER_SERVICE_UNAVAILABLE_CODE) {
```

sutaakar: Good point, updated.

PR #507, opened 2019-10-14
Check both HTTP and HTTPS routes for Workbench and Kie server deploym…

sutaakar:
…ents


sutaakar:
PR build failed due to different reason, merging for now.

PR #518, opened 2019-10-28
Look for image stream before setting operator image

radtriste:
This new code just take the image name defined and look if there is an existing image stream with the same name. If there is, then it replaces with the correct image repository and keeping the version (if none, set latest).

This should be backward compatible if we want to set an operator image directly.

### DIFF
--- a/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
+++ b/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
@@ -78,12 +78,12 @@ public abstract class OpenShiftOperatorScenario<T extends DeploymentScenario<T>>
 
     private void createCustomResourceDefinitionsInOpenShift(OpenShiftBinary adminBinary) {
         // TODO: Commented out due to UnrecognizedPropertyException, uncomment with raised Fabric8 version to check if it is fixed already.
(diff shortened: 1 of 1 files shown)


sutaakar:
Looks good to me.
Just one minor thing, can you please add there some comments explaining implementation logic?
It took me a bit of time to figure out all the options and implementation meanings.

radtriste:
Just added comments to the code. Please squash & merge if it is ok for you.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
```diff
@@ -104,7 +107,19 @@ private void createRoleBindingsInProject(Project project) {
...
+
+        // Get the operator image tag (composed of name + tag). 
+        // Retrieve the image name and see if it fits an image stream.
+        // If yes, then use the image stream image's name and same tag as defined (use latest if no tag).
+        // If not, use as it is as image name.
+        String operatorImageTag = OpenShiftOperatorConstants.getKieOperatorImageTag();
+        String[] split = operatorImageTag.split(":");
+        ImageStream operatorImageStream = project.getOpenShift().getImageStream(split[0]);
+        if (Objects.nonNull(operatorImageStream)) {
+            final String streamTag = split.length > 0 ? split[1] : "latest";
```
sutaakar:
```suggestion
            final String streamTag = split.length > 1 ? split[1] : "latest";
```

PR #524, opened 2019-11-04
[BAQE-1136] Use Kafka configuration directly from HA-CEP distribution

jiripetrlik:


framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
```diff
@@ -74,6 +72,10 @@
     private static final String SOURCES_FILE_ROLE_BINDING = "springboot/kubernetes/role-binding.yaml";
     private static final String SOURCES_FILE_SERVICE_ACCOUNT = "springboot/kubernetes/service-account.yaml";
 
+    private static final String SOURCES_KAFKA_TOPICS_FOLDER = "kafka-topics";
+    private static final String[] SOURCE_KAFKA_TOPIC_FILES =
+            new String[] {"control.yaml", "events.yaml", "kiesessioninfos.yaml", "snapshot.yaml"};
```
jakubschwan: Can you please add there comment from there are these topics from. It can not be clear in future.

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
```diff
@@ -168,50 +170,16 @@ private File downloadAndUnzipAMQStreams() {
...
-                .withReplicas(3)
-                .addConfigItem("retention.ms", "7200000")
-                .addConfigItem("segment.bytes", "1073741824")
-                .addConfigItem("cleanup.policy", "compact")
-                .addConfigItem("segment.ms", "100")
-                .addConfigItem("min.cleanable.dirty.ratio", "0.01")
-                .addConfigItem("delete.retention.ms", "100")
-                .build();
-        strimziOperator.createTopic(snapshotsTopic);
+        final File kafkaTopicsFolder = new File(OpenShiftConstants.getHaCepSources(), SOURCES_KAFKA_TOPICS_FOLDER);
```
jakubschwan: If we want to use files here, it would be better to include suffix Dir to the system property for HA CEP sources.
```suggestion
        final File kafkaTopicsFolder = new File(OpenShiftConstants.getHaCepSourcesDir(), SOURCES_KAFKA_TOPICS_FOLDER);
```

jiripetrlik:
@jakubschwan 
Thank you for comments. Both should be fixed.

PR #528, opened 2019-11-07
[BAQE-1150] Use Openshift build instead of pulling images for HA-CEP

jiripetrlik:
- Use Openshift build instead of pulling images, which is now recommended way in README
- Add some logging

### DIFF
--- a/framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
+++ b/framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
@@ -21,6 +21,7 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.URI;
 import java.net.URL;
 import java.security.KeyStore;
 import java.security.cert.CertificateFactory;
@@ -48,6 +49,8 @@ import org.kie.cloud.strimzi.deployment.StrimziOperatorDeployment;
 import org.kie.cloud.strimzi.deployment.ZookeeperDeployment;
 import org.kie.cloud.strimzi.resources.KafkaCluster;
 import org.kie.cloud.strimzi.resources.KafkaClusterBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class HACepScenarioImpl extends OpenShiftScenario<HACepScenario> implements HACepScenario {
 
@@ -67,6 +70,8 @@ public class HACepScenarioImpl extends OpenShiftScenario<HACepScenario> implemen
     private static final String SOURCES_FILE_ROLE = "springboot/kubernetes/role.yaml";
     private static final String SOURCES_FILE_ROLE_BINDING = "springboot/kubernetes/role-binding.yaml";
     private static final String SOURCES_FILE_SERVICE_ACCOUNT = "springboot/kubernetes/service-account.yaml";
+    private static final String SOURCES_FILE_HACEP_DEPLOYMENT = "springboot/kubernetes/deployment.yaml";
+    private static final String SOURCES_FILE_HACEP_SERVICE = "springboot/kubernetes/service.yaml";
 
     private static final String SOURCES_KAFKA_TOPICS_FOLDER = "kafka-topics";
     /* List of topics which needs to be created for HACEP. Topics are located in kafka-topics folder in HACEP
@@ -79,6 +84,8 @@ public class HACepScenarioImpl extends OpenShiftScenario<HACepScenario> implemen
 
     private StrimziOperator strimziOperator;
     private HACepDeployment haCepDeployment;
+
+    private static final Logger logger = LoggerFactory.getLogger(HACepScenarioImpl.class);
 
     @Override
     protected void deployKieDeployments() {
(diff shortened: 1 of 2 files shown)


framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
```diff
@@ -125,16 +135,31 @@ protected void deployKieDeployments() {
...
         project.createResourcesFromYamlAsAdmin(roleYamlFile.getAbsolutePath());
+        logger.info("Creating service account for HACEP from file: {}", serviceAccountYamlFile.getAbsolutePath());
         project.createResourcesFromYamlAsAdmin(serviceAccountYamlFile.getAbsolutePath());
+        logger.info("Creating role binding for HACEP from file: {}", roleBindingYamlFile.getAbsolutePath());
         project.createResourcesFromYamlAsAdmin(roleBindingYamlFile.getAbsolutePath());
 
-        project.createResourcesFromYamlAsAdmin(OpenShiftConstants.getHaCepResourcesList());
+        final String imageURL = buildHACEPImage();
+        final File haCepDeploymentYamlFile = new File(haCepSourcesDir, SOURCES_FILE_HACEP_DEPLOYMENT);
+        final String haCepDeploymentYaml = patchHACEPDeploymentFile(haCepDeploymentYamlFile, imageURL);
```
jakubschwan: I don't know how exactly the deployment file looks like, but couldn't be set the image url as parameter?

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
```diff
@@ -125,16 +135,31 @@ protected void deployKieDeployments() {
...
+
         haCepDeployment = new HACepDeploymentImpl(project);
+
+        logger.info("Exposing HACEP service as route");
         project.runOcCommandAsAdmin("expose", "service",
                                     ((HACepDeploymentImpl) haCepDeployment).getServiceName());
         haCepDeployment.waitForScale();
         try {
+            logger.info("Waiting 30 seconds for system to wake up (RHPAM-2380)");
             Thread.sleep(30000); // RHPAM-2380
```
jakubschwan: For what exactly we are waiting there? Could be there used SimpleWaiter or [waitUntilPodsAreReadyAndRunning](https://github.com/kiegroup/kie-cloud-tests/blob/master/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/deployment/OpenShiftDeployment.java#L160)

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
```diff
@@ -162,20 +189,52 @@ private File downloadAndUnzipAMQStreams() {
...
 
+    private String buildHACEPImage() {
+        project.runOcCommandAsAdmin("new-build", "--binary", "--strategy=docker",
+                                    "--name", "openshift-kie-springboot");
+        final File springModuleDir = new File(OpenShiftConstants.getHaCepSourcesDir(), "springboot");
+        logger.info("Building HA-CEP Spring boot image");
+        final String buildOutput = project.runOcCommandAsAdmin("start-build", "openshift-kie-springboot",
+                                    "--from-dir=" + springModuleDir.getAbsolutePath(), "--follow");
+        logger.info(buildOutput);
+        final String imageURL = project.getOpenShiftAdmin().getImageStream("openshift-kie-springboot").getStatus().getDockerImageRepository();
```
jakubschwan: Name imageURL for String can be confusing as we could expect URL object behind this name.
```suggestion
        final String dockerImageRepository = project.getOpenShiftAdmin().getImageStream("openshift-kie-springboot").getStatus().getDockerImageRepository();
```

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
```diff
@@ -162,20 +189,52 @@ private File downloadAndUnzipAMQStreams() {
...
+
+        return imageURL;
```
jakubschwan:
```suggestion
        return dockerImageRepository;
```

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
jiripetrlik: Seems that the old issue which caused this delay is no longer valid. Waiting can be removed.

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
jiripetrlik: Good point. Variable was renamed.

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
jiripetrlik: Same as above.

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/HACepScenarioImpl.java
jiripetrlik: I would prefer to keep it as it is. At least for now. It uses the same setup as other parts of HA-CEP automation, which expects unpacked folder with HACEP distribution in some folder. It is probably possible to allow urls, but this will affect much more parts of automation that it would be better to have a separate PR for it. Also there is Docker build running from folder with HACEP distribution using oc command.

jiripetrlik:
@jakubschwan 
Thank you for your review. PR should be fixed according to your review.

PR #537, opened 2019-12-02
[7.30.x] Add possibility to retrieve Maven repository from scenario

sutaakar:
Still doing some minor testing, but the PR should be moreless ready.

Should provide a bit cleaner implementation of MavenDeployer.
Should we still support using Maven repository deployed in dedicated namespace? Or rather completely drop it (as it most likely won't be used anyway)?

### DIFF
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/OpenShiftScenario.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/OpenShiftScenario.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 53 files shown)


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/api/scenario/KjarDeploymentScenarioListener.java
```diff
@@ -36,7 +36,9 @@ private KjarDeploymentScenarioListener(Kjar... kjars) {
 
     @Override
     public void beforeDeploymentStarted(T deploymentScenario) {
-        this.kjars.forEach(kjar -> KjarDeployer.create(kjar).deploy(deploymentScenario.getScenarioEnvironment()));
+        for (Kjar kjar : kjars) {
+            deploymentScenario.getMavenRepositoryDeployment().ifPresent(mavenDeployment -> KjarDeployer.create(kjar).deploy(mavenDeployment));
```
radtriste: Could put the for loop inside the Optional ?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/OpenShiftScenario.java
```diff
@@ -236,4 +242,12 @@ public void afterScenarioFinished(T deploymentScenario) {
 
         });
     }
+
+    @Override
+    public Optional<MavenRepositoryDeployment> getMavenRepositoryDeployment() {
```
radtriste: If we try to access the Maven Repository Deployment and there is none. That would mean there is no Kjar deployment ...
Shouldn't we raise an error instead ?

radtriste:
> Should we still support using Maven repository deployed in dedicated namespace? Or rather completely drop it (as it most likely won't be used anyway)?

I would say that if we go with each test (which are independent) have its own Nexus, we don't need global Maven anymore or ?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/scenario/OpenShiftScenario.java
sutaakar: Well, that is a question.
I think it depends on the fact if the information about whether Maven repository is deployed is relevant to user. If user calls this method to figure out whether the Maven repository is deployed or rather just want to get it.

Well, for our case I suppose that when user calls this method then he needs to use Maven repository. Will adjust the implementation to throw an exception.

sutaakar:
@radtriste All issues should be addressed.

PR #543, opened 2019-12-16
Scale immutable Kie server to just one instance

sutaakar:
As OptaPlanner tests don't correctly work on multiple Kie server instances.


jakubschwan:
@sutaakar Changes are done in scenario. Do we want to apply this changes for all immutable tests or only for OptaPlanner tests?

sutaakar:
Can keep it for all tests to be consistent.

PR #561, opened 2020-01-13
Change log message layout to include class, method and line

Sgitario:
Adding this information in the log message might help to troubleshoot issues when reading the log messages.


sutaakar:
Is it ready for review and merge?
Or there is other work expected?

PR #570, opened 2020-01-23
Master test runners

Sgitario:
Testing one scenario, I saw that the properties inside the template-xxx.properties file were not working fine using TestRunners. I fixed this by moving some dependencies.

Also, since we're in the master, I upgraded the version to master and 77.


sutaakar:
@Sgitario Please squash and rebase

PR #577, opened 2020-01-31
[7.30.x] Implement test for external JMS signal

sutaakar:


framework-cloud/framework-openshift-apb/src/main/java/org/kie/cloud/openshift/scenario/builder/WorkbenchRuntimeSmartRouterImmutableKieServerAmqWithDatabaseScenarioBuilderApb.java
```diff
@@ -146,4 +146,9 @@ public WorkbenchRuntimeSmartRouterImmutableKieServerAmqWithDatabaseScenarioBuild
         return this;
     }
 
+    @Override
+    public WorkbenchRuntimeSmartRouterImmutableKieServerAmqWithDatabaseScenarioBuilder enableExternalJmsSignalQueue(String queueName) {
+        // TODO Auto-generated method stub
```
Sgitario: I think this comment stands for something that it's pending to be added. If APB does not support this feature, I think it's better to throw an exception for this purpose.

PR #590, opened 2020-03-01
[BAQE-1295] Upgrade versions from 7.7 to 7.8

Sgitario:


Sgitario:
@jakubschwan can you take a look into this PR? Thanks.

PR #591, opened 2020-03-03
BAQE-1261: Create dedicated GOGS instance for each Openshift project

Sgitario:
The idea was to create a dedicated GOGS instance as GIT Provider that lives in the same namespace as the running project test. This way we don't depend on third services to ensure the tests. 

Some problems I got during the development of this task were:
- Tests were using the GIT provider in a different test lifecycle (sometimes before the project even existed)
- We still want the ability to use the external GOGS instance

To manage the above challenges, I had to refactor all the tests to use this GIT instance the same way. Also, I followed the same approach as the LDAP deployment. Mainly, the tests will:

1. First create a TEST repository name. (a random string only). Example:

```java
private static final String REPOSITORY_NAME = generateNameWithPrefix(ProcessFailoverIntegrationTest.class.getSimpleName());
```

2.- Configure the GIT settings:

```java
.withGitSettings(GitSettings.fromProperties()
                                                                    .withRepository(REPOSITORY_NAME,
ProcessFailoverIntegrationTest.class.getResource(PROJECT_SOURCE_FOLDER + "/" + DEFINITION_PROJECT_NAME).getFile()))
```

Similarly as the LDAP settings.

3.- Now, when using the GIT provider, we need to use the deployment scenario:

```java
GitProvider gitProvider = deploymentScenario.getGitProvider()
```

3.- Finally, we still need to delete the repository in the after stage since by configuration, the GIT instance might be still using an external service:

```java
@After
    public void tearDown() {
        GitUtils.deleteGitRepository(REPOSITORY_NAME, deploymentScenario);
    }
```

### DIFF
--- a/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/KieDeploymentScenario.java
+++ b/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/KieDeploymentScenario.java
@@ -17,10 +17,11 @@ package org.kie.cloud.api.scenario;
 
 import java.util.List;
 
+import org.kie.cloud.api.deployment.ControllerDeployment;
 import org.kie.cloud.api.deployment.KieServerDeployment;
 import org.kie.cloud.api.deployment.SmartRouterDeployment;
 import org.kie.cloud.api.deployment.WorkbenchDeployment;
-import org.kie.cloud.api.deployment.ControllerDeployment;
+import org.kie.cloud.api.git.GitProvider;
 
 /**
  * Cloud deployment scenario representation.
@@ -62,4 +63,13 @@ public interface KieDeploymentScenario<T extends DeploymentScenario<T>> extends 
      * @see ControllerDeployment
      */
     List<ControllerDeployment> getControllerDeployments();
+
+    /**
+     * Return GIT provider.
+     *
+     * @return GIT provider.
+     */
+    default GitProvider getGitProvider() {
+        throw new RuntimeException("No GIT provider configured for this scenario");
+    }
 }
--- a/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/builder/ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder.java
+++ b/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/builder/ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder.java
@@ -16,6 +16,7 @@
 package org.kie.cloud.api.scenario.builder;
 
 import org.kie.cloud.api.scenario.ClusteredWorkbenchKieServerDatabasePersistentScenario;
+import org.kie.cloud.api.settings.GitSettings;
 import org.kie.cloud.api.settings.LdapSettings;
 
 public interface ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder extends DeploymentScenarioBuilder<ClusteredWorkbenchKieServerDatabasePersistentScenario> {
@@ -44,11 +45,20 @@ public interface ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder ex
     ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder deploySso();
 
     /**
-     * Return setup builder with an external LDAP.
+     * Return setup builder with additional GIT settings.
      *
-     * @param ldapSettings configuration of LDAP represented by a class.
+     * @param gitSettings settings configuration of GIT
      * @return Builder
      */
+    ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder withGitSettings(GitSettings gitSettings);
+
+    /**
+    *
+    * Return setup builder with an external LDAP.
+    *
+    * @param ldapSettings configuration of LDAP represented by a class.
+    * @return Builder
+    */
     ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder withExternalLdap(LdapSettings ldapSettings);
 
     /**
@@ -100,4 +110,5 @@ public interface ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder ex
      * @return Builder with configured memory limit.
      */
     ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder withWorkbenchMemoryLimit(String limit);
+
 }
--- a/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/builder/ImmutableKieServerAmqScenarioBuilder.java
+++ b/framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/builder/ImmutableKieServerAmqScenarioBuilder.java
@@ -16,6 +16,7 @@
 package org.kie.cloud.api.scenario.builder;
 
 import org.kie.cloud.api.scenario.ImmutableKieServerAmqScenario;
+import org.kie.cloud.api.settings.GitSettings;
 import org.kie.cloud.api.settings.LdapSettings;
 
 public interface ImmutableKieServerAmqScenarioBuilder extends KieDeploymentScenarioBuilder<ImmutableKieServerAmqScenarioBuilder, ImmutableKieServerAmqScenario> {
@@ -46,24 +47,22 @@ public interface ImmutableKieServerAmqScenarioBuilder extends KieDeploymentScena
     /**
      * Return configured builder with Source location
      *
-     * @param gitRepoUrl Repository url.
      * @param gitReference Repository reference (branch/tag). E.g. 'master'.
      * @param gitContextDir Repository context (location of pom file).
      * @return Builder
      */
-    ImmutableKieServerAmqScenarioBuilder withSourceLocation(String gitRepoUrl, String gitReference, String gitContextDir);
+    ImmutableKieServerAmqScenarioBuilder withSourceLocation(String gitReference, String gitContextDir);
 
     /**
      * Return configured builder with Source location
      *
-     * @param gitRepoUrl Repository url.
      * @param gitReference Repository reference (branch/tag). E.g. 'master'.
      * @param gitContextDir Repository context (location of pom file).
      * @param artifactDirs Directories containing built kjars, separated by
      * commas. For example "usertask-project/target,signaltask-project/target".
      * @return Builder
      */
-    ImmutableKieServerAmqScenarioBuilder withSourceLocation(String gitRepoUrl, String gitReference, String gitContextDir, String artifactDirs);
+    ImmutableKieServerAmqScenarioBuilder withSourceLocation(String gitReference, String gitContextDir, String artifactDirs);
 
     /**
      * Return setup builder with additional configuration for SSO deployment.
@@ -114,4 +113,12 @@ public interface ImmutableKieServerAmqScenarioBuilder extends KieDeploymentScena
      * @return Builder with configured internal ldap.
      */
     ImmutableKieServerAmqScenarioBuilder withInternalLdap(LdapSettings ldapSettings);
+
+    /**
+     * Return setup builder with additional GIT settings.
+     *
+     * @param gitSettings settings configuration of GIT
+     * @return Builder
+     */
+    ImmutableKieServerAmqScenarioBuilder withGitSettings(GitSettings gitSettings);
 }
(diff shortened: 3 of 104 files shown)


framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/KieDeploymentScenario.java
```diff
@@ -62,4 +64,13 @@
      * @see ControllerDeployment
      */
     List<ControllerDeployment> getControllerDeployments();
+
+    /**
+     * Return GIT provider.
+     *
+     * @return GIT provider.
+     */
+    default Optional<GitProvider> getGitProvider() {
```
sutaakar: What is the motivation for using Optional here? Is empty git provider a valid scenario?

Sgitario: I thought GitProvider is not always required, that's why the optional. If it's always required, we could remove the Optional and return the new GitProviderService().buildGitProvider() by default?

sutaakar: > If it's always required, we could remove the Optional and return the new GitProviderService().buildGitProvider() by default?

Sounds good to me.
There has been a discussion about using Optional for such cases some time ago (when Maven was refactored). The result is that Optional is not meaningful in this case. 

The reasons behind were that this framework is not intended to be used in other framework or software, but rather by tests themselves. If the test calls getGitProvider then it expect to get GitProvider instance. Returning empty value is not a valid option for test. Therefore if the provider is not available for some reason then it is better to throw an exception (fail fast).

Sgitario: It makes sense to me. If caller calls to getGitProvider() it should fail if the scenario does not support it.
Will change it soon. 
Thanks,

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/settings/GitSettings.java
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * Class to configure the GIT instance to be used among the tests.
+ *
+ * Currently, there is only one way to create the GitSettings via properties (see method .fromProperties()).
+ */
+public class GitSettings {
+
+    public static final String GIT_PROVIDER = "git.provider";
+
+    private final String provider;
```
sutaakar: This property doesn't seem to be used anywhere.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/Git.java
```diff
@@ -0,0 +1,71 @@
...
+public class Git {
+
+    private static final String GOGS = "Gogs";
+    private static final Map<String, Function<Project, GitProvider>> SCENARIO_PROVIDERS = Collections.singletonMap(GOGS, deployGogs());
+
+    private static GitProvider PROVIDER;
+
+    public static synchronized GitProvider getProvider(Project project, GitSettings settings) {
+        if (PROVIDER == null) {
+            PROVIDER = Optional.ofNullable(SCENARIO_PROVIDERS.get(GitConstants.getGitProvider()))
```
sutaakar: This functionality is supposed to handle internal or external GIT provider initialization, right?

Does it have sense to support the external GIT provider anyway? AFAIK we can use just internal one, it will simplify implementation and makes our life easier.

sutaakar: As side note, if I understand the code properly. 
The getProvider() method is used by all scenarios to get GIT provider. However the PROVIDER is a static field, so all scenarios are pointing to same GIT provider now, right?

Sgitario: I think this is a leftover from the old approach. Will remove it. Thanks for spotting!

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/KieDeploymentScenario.java
Sgitario: Done.

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/KieDeploymentScenario.java
```diff
@@ -62,4 +64,13 @@
...
+        return Optional.empty();
+    };
```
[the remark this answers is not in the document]
Sgitario: Done

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/builder/ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder.java
```diff
@@ -44,11 +45,20 @@
     ClusteredWorkbenchKieServerDatabasePersistentScenarioBuilder deploySso();
 
     /**
-     * Return setup builder with an external LDAP.
+     * Return setup builder with additional GIT settings.
      *
-     * @param ldapSettings configuration of LDAP represented by a class.
+     * @param git settings
```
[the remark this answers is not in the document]
Sgitario: Done

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/scenario/builder/ImmutableKieServerAmqScenarioBuilder.java
```diff
@@ -114,4 +113,12 @@
      * @return Builder with configured internal ldap.
      */
     ImmutableKieServerAmqScenarioBuilder withInternalLdap(LdapSettings ldapSettings);
+
+    /**
+     * Return setup builder with additional GIT settings.
+     *
+     * @param git settings
```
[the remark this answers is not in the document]
Sgitario: done

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/settings/GitSettings.java
Sgitario: It should be used in org.kie.cloud.openshift.util.Git. Fixed.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/Git.java
Sgitario: Removed PROVIDER field as it was completely unnecessary. 
I didn't remove the external GIT configuration, but I agree with you: if it's not necessary any longer, we should get rid of it.

framework-cloud/framework-openshift/pom.xml
```diff
@@ -22,6 +22,10 @@
       <groupId>org.kie.cloud</groupId>
       <artifactId>framework-cloud-common</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.kie.cloud</groupId>
+      <artifactId>framework-git</artifactId>
```
sutaakar: My intention was to keep framework-openshift independent from framework-git, allowing test modules to import just parts of the framework which are needed.
Can you remove the git dependency from here and keep it rather in test module?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/Git.java
```diff
@@ -0,0 +1,66 @@
...
+ * Once the GitProvider is created, it will raise a repository loaded event (using the Observer pattern), so external configuration
+ * can be updated transparently.
+ *
+ */
+public class Git {
+
+    private static final String GOGS = "Gogs";
+    private static final Map<String, Function<Project, GitProvider>> SCENARIO_PROVIDERS = Collections.singletonMap(GOGS, deployGogs());
+
+    public static synchronized GitProvider getProvider(Project project, GitSettings settings) {
```
sutaakar: Can you please rename the method to createProvider, to reflect what is the method actually doing? :)

framework-cloud/framework-openshift/pom.xml
Sgitario: I provided a solution to remove this dependency using ServiceLoader framework. 

The latest decisions have been made:

- Update GitProviderService to use the service loader too. This service is still in use in the performance tests and test initializer. 
- Move Gogs deployed into the test-remote-cloud and change it to implement the GitProviderFactory instead of being an utility. 
- Since the Gogs provider factory needs the Project to deploy into Openshift, I created a new interface ProjectInitializer to initialize all the services that need this Project instance. This might be used for other service as well.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/util/Git.java
Sgitario: Done

pom.xml
```diff
@@ -24,7 +24,7 @@
 
   <properties>
     <rhba.images.version.prefix>77</rhba.images.version.prefix>
-    <rhba.images.branch>master</rhba.images.branch>
+    <rhba.images.branch>7.7.x</rhba.images.branch>
```
jakubschwan: please set it back to master as this PR is to master branch, you probably changed that for testing purposes

test-cloud/test-cloud-remote/src/test/resources/test.properties
```diff
@@ -26,5 +26,5 @@ amq.password=usetheforce123@
 kie.artifact.version=please-define-kie-artifact-version
 kjars.build.settings.xml=src/test/resources/settings.xml
 
-rhba.images.branch=master
+rhba.images.branch=7.7.x
```
jakubschwan: same here with setting it back to master

framework-cloud/framework-cloud-api/src/main/java/org/kie/cloud/api/settings/GitSettings.java
```diff
@@ -0,0 +1,73 @@
...
+        return repositoryLoadedActions;
+    }
+
+    public GitSettings withRepository(String repositoryName, String repositoryPath) {
+        this.repositoryName = repositoryName;
+        this.repositoryPath = repositoryPath;
+        return this;
+    }
+
+    public void addOnRepositoryLoaded(Consumer<String> repositoryLoadedAction) {
```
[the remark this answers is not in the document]
Sgitario: Done

pom.xml
Sgitario: My bad. Reverted

test-cloud/test-cloud-remote/src/test/resources/test.properties
Sgitario: My bad. Reverted

framework-cloud/framework-openshift-templates/src/main/java/org/kie/cloud/openshift/scenario/builder/ImmutableKieServerScenarioBuilderImpl.java
```diff
@@ -64,30 +66,36 @@ public ImmutableKieServerScenarioBuilder withContainerDeployment(String kieConta
     }
 
     @Override
-    public ImmutableKieServerScenarioBuilder withSourceLocation(String gitRepoUrl, String gitReference, String gitContextDir) {
-        envVariables.put(OpenShiftTemplateConstants.SOURCE_REPOSITORY_URL, gitRepoUrl);
-        envVariables.put(OpenShiftTemplateConstants.SOURCE_REPOSITORY_REF, gitReference);
-        envVariables.put(OpenShiftTemplateConstants.CONTEXT_DIR, gitContextDir);
+    public ImmutableKieServerScenarioBuilder withSourceLocation(String gitReference, String gitContextDir) {
```
sutaakar: The gitReference and gitContextDir doesn't seem to be used here? Did you accidentally deleted env varible configuration from above?

Sgitario: Thanks for spotting this!
Fixed

sutaakar:
@Sgitario Can you please squash and rebase to the master branch, resolving possible conflicts?

Sgitario:
Conflicts resolved.

PR #599, opened 2020-03-11
[BAQE-1315] Adjust certificate setup also for HA-CEP

jiripetrlik:


test-cloud/test-cloud-ha-cep/pom.xml
```diff
@@ -123,6 +123,16 @@
         </plugin>
       </plugins>
     </pluginManagement>
+    <plugins>
```
sutaakar: Would it have sense to declare these plugins directly in test-cloud?

jiripetrlik: Good point. I've moved these plugins to test-cloud module.

PR #600, opened 2020-03-11
BAQE-1332: Add External DB drivers test suite for Operators

Sgitario:
In order to avoid regression issues, I run the DB job for templates: https://rhba-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/jcarvaja/job/jcarvaja-openshift-db-templates/7/

And I checked the DB job for operators which now works fine (in the same engines as in templates): https://rhba-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/jcarvaja/job/jcarvaja-openshift-db-operator/40/

### DIFF
--- a/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/KieServerWithExternalDatabaseScenarioImpl.java
+++ b/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/KieServerWithExternalDatabaseScenarioImpl.java
@@ -62,7 +62,7 @@ public class KieServerWithExternalDatabaseScenarioImpl extends OpenShiftOperator
 
         dockerDeployment = DockerRegistryDeployer.deploy(project);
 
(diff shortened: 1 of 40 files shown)


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/KieServerWithExternalDatabaseScenarioImpl.java
```diff
@@ -0,0 +1,156 @@
...
+        super(kieApp);
+    }
+
+    @Override
+    protected void deployCustomResource() {
+        OperatorExternalDatabase externalDatabase = OperatorExternalDatabaseProvider.getExternalDatabase();
+
+        dockerDeployment = DockerRegistryDeployer.deploy(project);
+
+        // Create image stream from external image with driver and reference it for template
```
sutaakar:
```suggestion
        // Create image stream from external image with driver and reference it for custom resource
```

PR #601, opened 2020-03-12
BAQE-1327 Fix WorkbenchPersistenceIntegrationTest tests in RHDM/RHPAM

Sgitario:


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/persistence/WorkbenchPersistenceIntegrationTest.java
```diff
@@ -185,9 +189,11 @@ private void assertSpaceAndProjectExists(String spaceName, String projectName) {
     }
 
     private void scaleToZeroAndToOne(Deployment deployment) {
-        deployment.scale(0);
-        deployment.waitForScale();
-        deployment.scale(1);
-        deployment.waitForScale();
+        waitUntilKieServerIsConnectedAfterAction(kieServerDeployment, () -> {
```
sutaakar: This may break down again as soon as we enable operator for this test scenario, as operator will use OpenShiftStartupStrategy.
Btw why does it matter if Kie server reconnects or not? AFAIK the test are making sure that persistence of Workbench works, if Workbench is restarted it should already contain all the needed informations.

Sgitario: At the moment, it's also checking the server instances, see verifyOneServerTemplateWithContainer method.

sutaakar: Oh, ok.
Still, would it be possible to have a functionality checking that Kie server is connected into the Workbench regardless of the startup strategy?

Sgitario: Another option is using the Awailability framework to check the server templates. 
In verifyOneServerTemplateWithContainer, something like:

awaits().until(kieControllerClient.listServerTemplates().getServerTemplates().size() > 0);

It's pretty the same we're doing right now to check the logs, but this framework is fully integrated with JUnit and as far I could see, the version 3.0.0 is already registered in the kie-parent module. 

What do you think about?

sutaakar: Sounds good to me.

PR #603, opened 2020-03-16
BAQE-1327: Fix WorkbenchPersistenceIntegrationTest tests in RHDM/RHPAM

Sgitario:
Actions:
- Wait until KIE Server is connected back to Business Central after Business Central restart
- Use Awaitility framework to check list of server instances

The Awaitility framework is fully well integrated with JUnit and this version (3.0.0 a bit old) was already managed by kie-parent.

RHPAM Jenkins Job that confirms this test is now working fine: https://rhba-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/jcarvaja/job/RHPAM-7.7-verification-openshift-single/

For RHDM: https://rhba-qe-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/custom/job/jcarvaja/job/RHDM-7.7-verification-openshift-single/

### DIFF
--- a/test-cloud/test-cloud-remote/pom.xml
+++ b/test-cloud/test-cloud-remote/pom.xml
@@ -72,6 +72,12 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>compile</scope>
+    </dependency>
 
     <!-- XTF dependencies -->
     <dependency>
--- a/test-cloud/test-cloud-remote/src/main/java/org/kie/cloud/utils/AwaitilityUtils.java
+++ b/test-cloud/test-cloud-remote/src/main/java/org/kie/cloud/utils/AwaitilityUtils.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.kie.cloud.utils;
+
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+
+import org.awaitility.Awaitility;
+import org.hamcrest.Matcher;
+import org.hamcrest.Matchers;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Awaitility utils to make a long or repeatable operation.
+ */
+public class AwaitilityUtils {
+
+    private static final Logger logger = LoggerFactory.getLogger(AwaitilityUtils.class);
+
+    private AwaitilityUtils() {
+
+    }
+
+    /**
+     * Wait until supplier returns a not null instance.
+     *
+     * @param supplier method to return the instance.
+     * @return the non null instance.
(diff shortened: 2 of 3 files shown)


test-cloud/test-cloud-remote/src/main/java/org/kie/cloud/utils/AwaitilityUtils.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param supplier method to return the instance.
+     * @return the non empty array.
+     */
+    public static final <T> T[] untilIsNotEmpty(Supplier<T[]> supplier) {
+        return until(supplier, Matchers.arrayWithSize(Matchers.greaterThan(0)));
+    }
+
+    private static final <T> T until(Supplier<T> supplier, Matcher<T> matcher) {
+        return Awaitility.await()
+                         .pollInterval(30, TimeUnit.SECONDS)
```
sutaakar: That is quite long poll interval, could you make it shorter, for example 5 seconds? Or maybe even less?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/persistence/WorkbenchPersistenceIntegrationTest.java
```diff
@@ -164,11 +164,14 @@ public void testWorkbenchProjectPersistence() {
         assertThat(containersResponse.getResult().getContainers().get(0).getContainerId()).isEqualTo(CONTAINER_ID);
     }
 
-    private void verifyOneServerTemplateWithContainer(String containerId) {
-        ServerTemplateList serverTemplates = kieControllerClient.listServerTemplates();
-        assertThat(serverTemplates.getServerTemplates()).as("Number of server templates differ.").hasSize(1);
+    private ServerTemplate[] verifyOneServerTemplate() {
+        ServerTemplate[] serverTemplates = AwaitilityUtils.untilIsNotEmpty(() -> kieControllerClient.listServerTemplates().getServerTemplates());
+        assertThat(serverTemplates).as("Number of server templates differ.").hasSize(1);
```
sutaakar: Right now this method does two things, returns list of server templates and at the same time checking that there is just one template.
Can you move this assertion to verifyOneServerTemplateWithContainer(), possibly renaming this method to getServerTemplates?

test-cloud/test-cloud-remote/src/main/java/org/kie/cloud/utils/AwaitilityUtils.java
Sgitario: Agree. Done.

test-cloud/test-cloud-remote/pom.xml
```diff
@@ -72,6 +72,12 @@
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
     </dependency>
+    
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+      <scope>compile</scope>
```
jakubschwan: just minor thing, do we need there scope?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/persistence/WorkbenchPersistenceIntegrationTest.java
Sgitario: I did this way because I needed to use within the method scaleToZeroAndToOne. Instead of the changes you suggested, I did:
- The method verifyOneServerTemplate won't return anything: so it will only do one thing as good practice.
- The method verifyOneServerTemplateWithContainer won't call to verifyOneServerTemplate (since it was already called by scaleToZeroAndToOne). To make this method safer, I provided a new method in Awaitibility called untilAsserted, so we can easily assert things using this framework.

test-cloud/test-cloud-remote/pom.xml
Sgitario: Yes, because the kie-parent is marked this dependency with the scope "test", and we created an utility wrapper in the sources.

sutaakar:
Looks good to me.

PR #605, opened 2020-03-16
BAQE-1325: Mark tests that require missing properties as Ignored in RHDM

Sgitario:
These tests do not support DROOLS, so in order to avoid setup the environment, we can quickly skip those when running in RHDM.

### DIFF
--- a/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/ldap/KieServerWithMySqlLdapIntegrationTest.java
+++ b/test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/ldap/KieServerWithMySqlLdapIntegrationTest.java
@@ -85,7 +85,6 @@ public class KieServerWithMySqlLdapIntegrationTest extends AbstractCloudIntegrat
     }
 
     @Test
-    @Category(JBPMOnly.class)
     public void testProcessFromMavenRepo() {
         processTestProvider.testDeployFromKieServerAndExecuteProcesses(deploymentScenario.getKieServerDeployment());
     }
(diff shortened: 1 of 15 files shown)


Sgitario:
Does it make sense to apply these changes into the 7.33.x branch too?

sutaakar:
It is not needed, so up to you.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/smoke/KieServerWithBuiltKjarIntegrationTest.java
```diff
@@ -51,7 +52,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@Category(Smoke.class)
+@Category({Smoke.class, JBPMOnly.class})
```
jakubschwan: This test class is valid for drools, please remove JBPMOnly category.
See configuration for drools tests in this class https://github.com/kiegroup/kie-cloud-tests/pull/605/files#diff-529defef45ab113c32c0e3b7339f933eR109-R118

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/ldap/KieServerWithMySqlLdapIntegrationTest.java
```diff
@@ -31,6 +31,7 @@
 import org.kie.cloud.tests.common.ScenarioDeployer;
 import org.kie.cloud.tests.common.client.util.LdapSettingsConstants;
 
+@Category(JBPMOnly.class)
```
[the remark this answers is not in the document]
Sgitario: Done.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/smoke/KieServerWithBuiltKjarIntegrationTest.java
Sgitario: Thanks for spotting Jakub.
So, from the three scenarios in this test suite, two are not valid for drools. In order to avoid to initialize these two scenarios in Openshift, I will follow the same approach as other scenario builders: check the scenario in the builder constructor with:

```java
isScenarioValid();
```

This method throws an UnsupportedOperationException to skip the scenario.

Moreover, I created a new utility class ScenarioValidations.java to have methods to verify whether it's running on either drools or jbpm, and avoid that the scenario builders have to implement this verification by their own.

PR #625, opened 2020-04-29
BAQE-1385: Missing SSO configuration in WorkbenchKieServerPersistentS…

Sgitario:
…cenarioImpl
JIRA Ticket: https://issues.redhat.com/browse/BAQE-1385
Description: Add missing SSO configuration in this scenario.
It's not needed in Operator.

This is caused by https://issues.redhat.com/browse/BAQE-1359.


Sgitario:
After this change, the test KieServerS2iWithSsoDroolsIntegrationTest is working fine.

PR #626, opened 2020-04-29
BAQE-1386: Make persistent tests more resilient

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/BAQE-1386
Description: 
What this test PersistenceTestProvider is doing is basically:
1.- Deploy container into Kie Server
2.- Restart Workbench
3.- Verify Workbench is still connected to Kie Server with the expected containers

The problem is that when restarting the Workbench, it might take some time to reconnect with the Kie Servers.

Solution:
My change is about to make the step 3 to wait until the asserts are working fine. Another possibility would be to wait for the re connection between kie-server and workbench and then run the step 3.


Sgitario:
@jakubschwan can you have a look into this PR so it can be merged? Thanks.

PR #628, opened 2020-05-05
BAQE-1393: Update workarounds to the new issue and delete RHPAM-2830

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/BAQE-1393
Update workarounds as RHPAM-2830 was resolved and the issue with the production template has been separated into RHPAM-2919


Sgitario:
I have run the tests that use the production environment and they are working using this workaround.

PR #631, opened 2020-05-19
BAQE-1316: Increase RHSSO to 7.4.0 in kie-cloud-tests

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/BAQE-1316
Description: Use latest 7.4.0 for cloud tests

I've ran a few tests using the latest SSO and they worked as expected.
I didn't need to upgrade the Maven dependencies for keycloak.

### DIFF
--- a/framework-cloud/framework-openshift/pom.xml
+++ b/framework-cloud/framework-openshift/pom.xml
@@ -44,10 +44,6 @@
     <!-- RH SSO dependencies  -->
     <dependency>
       <groupId>org.keycloak</groupId>
-      <artifactId>keycloak-core</artifactId>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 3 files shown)


framework-cloud/framework-openshift/pom.xml
```diff
@@ -42,10 +42,6 @@
     </dependency>
 
     <!-- RH SSO dependencies  -->
-    <dependency>
-      <groupId>org.keycloak</groupId>
-      <artifactId>keycloak-core</artifactId>
-    </dependency>
```
Sgitario, line 48: This dependency was not necessary as it's already part of keycloak-admin-client.

Sgitario:
@sutaakar or @jakubschwan Can this PR be merged already?

PR #632, opened 2020-05-28
BAQE-1421: Fix External Database tests that need download drivers

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/BAQE-1421
Description:
- Add new BUILD_ENGINE to the cekit command (by default, it uses podman and the jenkins node does not have it installed)
- Change the image name (it didn't match with the one in the templates.zip...)
- Fixed the repository (see more about this issue in https://issues.redhat.com/browse/RHPAM-2948)
- Removed unused "kie.jdbc.driver.binary.url" parameter
- Download driver only for propietary databases (oracle and sybase)

### DIFF
--- a/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/database/driver/AbstractExternalDriver.java
+++ b/framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/database/driver/AbstractExternalDriver.java
@@ -16,12 +16,13 @@
 package org.kie.cloud.openshift.database.driver;
 
 import java.net.URL;
+import java.util.Optional;
 
 public abstract class AbstractExternalDriver implements ExternalDriver {
 
     @Override
     public String getSourceDockerTag() {
-        return "kiegroup/" + getImageName() + ":" + getImageVersion();
+        return "quay.io/kiegroup/" + getImageName() + ":" + getImageVersion();
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


Sgitario:
Depends on https://github.com/jboss-integration/bxms-qe-tests/pull/3297

Sgitario:
@jakubschwan and @sutaakar can you take a look into this PR? Thanks.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/database/driver/AbstractExternalDriver.java
```diff
@@ -31,6 +31,11 @@ public String getTargetDockerTag(URL dockerUrl) {
 
     @Override
     public String getCekitImageBuildCommand() {
-        return "make build " + getName();
+        return "make build " + getName() + " BUILD_ENGINE=docker";
+    }
+
+    @Override
+    public String getJdbcDriverUrl() {
```
sutaakar, line 38: What about making the response Optional? That would reflect the fact that the driver URL doesn't have to be specified.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/database/driver/Db2ExternalDriver.java
```diff
@@ -31,4 +33,9 @@ public String getImageName() {
     public String getImageVersion() {
         return "11.1.4.4";
     }
+
+    @Override
+    public String getJdbcDriverUrl() {
```
sutaakar, line 38: DB2 can be built without external driver file. We can use it.

Sgitario: Thanks for spotting! Will revert this change.

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/database/driver/AbstractExternalDriver.java
Sgitario: Changed

PR #633, opened 2020-06-02
RHPAM-2762 & RHPAM-2777 add tests to verify issues fixImplement test …

jakubschwan:
…scenarios using persistent test scenarioold test provider and change user impl will be removed in following commitClean framework and tests


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/WorkbenchKieServerPersistentScenarioBuilderImpl.java
```diff
@@ -68,11 +68,9 @@ public WorkbenchKieServerPersistentScenarioBuilderImpl() {
         kieApp.getSpec().setCommonConfig(commonConfig);
 
         Server server = new Server();
-        server.addEnvs(authenticationEnvVars);
         kieApp.getSpec().getObjects().addServer(server);
 
         Console console = new Console();
-        console.addEnvs(authenticationEnvVars);
```
Sgitario, line 75: I guess these changes are intentional. If so, the authenticationEnvVars list is no longer used and it can be removed.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/WorkbenchKieServerPersistentScenarioImpl.java
```diff
@@ -89,24 +94,29 @@ protected void deployCustomResource() {
...
-        registerTrustedSecret(kieApp.getSpec().getObjects().getConsole());
-
+        if (!isCustomTrustedSecretRegistered(kieApp.getSpec().getObjects().getConsole())) {
+            registerTrustedSecret(kieApp.getSpec().getObjects().getConsole());
+        }
         for (Server server : kieApp.getSpec().getObjects().getServers()) {
-            registerTrustedSecret(server);
+            if (!isCustomTrustedSecretRegistered(server)) {
+                registerTrustedSecret(server);
+            }
```
Sgitario, line 103: Why do we need to check whether the ssl is configured? The values will remain the same, so it should not matter if we set the environment variables again and they won't change in the second execution, right?

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/WorkbenchKieServerPersistentScenarioImpl.java
```diff
@@ -179,4 +193,36 @@ public SsoDeployment getSsoDeployment() {
...
+            kieServerDeployment.waitForScale();
+        } else{
+            throw new RuntimeException("Application is not ready for Username and password change. Please check first that application is ready.");
+        }
+    }
+
+    private boolean areInstancesDeleted(Collection<String> oldInstancesNames) {
+        return project.getOpenShift().getPods().stream().map(Pod::getMetadata)
+                                                        .map(ObjectMeta::getName)
+                                                        .noneMatch(oldInstancesNames::contains);
```
Sgitario, line 226: Question about this: so when we change the username and password, the deployment is restarted. This means that the pods are regenerated completely? So the restart count from the container statuses remains to zero? Asking whether we can use the OpenShiftWaiters.havePodsBeenRestarted for this.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/WorkbenchKieServerPersistentScenarioImpl.java
jakubschwan: Yes, the value remain same but once when ssl is configured, I think there is no need to creating it again. If there is existing configuration, we should use it. It's also closer to manual update, when is KieApp updated manually we do not created the custom secret again.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/WorkbenchKieServerPersistentScenarioImpl.java
jakubschwan: Could be an option, but for tests we are using TestProviders, so the scenario is started once, but there are several changes and restarts. I think that this check, when we are waiting for old pod removing is better.

Sgitario: Ok ok

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/WorkbenchKieServerPersistentScenarioImpl.java
Sgitario: Ok, I see the problem. The "env" is a field, if we don't check this, we'll be adding the same value several times. 
Maybe in order to avoid having the same hard coded values in difference places (example: "HTTPS_NAME" or "HTTPS_PASSWORD"), what about adding a transient method in the Server.java to add the env if it does not exist?

```java
    @Transient
    public void addEnvIfNotSet(String name, String value) {
        if (Stream.of(getEnv()).map(Env::getName).noneMatch(envName -> StringUtils.equals(envName, name))) {
            addEnv(new Env(name, value));
        }
    }
```

Then, it would be only about to replace the existing "addEnv" method to "addEnvIfNotSet" in OpenShiftOperatorScenario. We would not need to worry about having new methods to check existing envs or deal with duplicated hardcoded values.
What do you think?

PR #634, opened 2020-06-03
BAQE-1421 Fix sybase driver and revert BUILD_ENGINE param

Sgitario:
JIRA Ticket: https://issues.redhat.com/browse/BAQE-1421
Description: 
- Sybase needs the KIE_SERVER_EXTERNALDB_URL property to be set (the default jdbc url is not working)
- The BUILD_ENGINE parameter was changed to Docker as part of RHPAM-2948


Sgitario:
I ran the tests and they worked fine.

PR #635, opened 2020-06-03
BAQE-1426: Fix Operator Deployment

Sgitario:
JIRA: https://issues.redhat.com/browse/BAQE-1426
Description: 

As part of RHPAM-2684, the operator is adding some console links to the KieApp object.
The problem is that the test framework is not importing the credentials, so the operator can generate these links. Concretely, the credentials are:

cluster_role_binding.yaml
cluster_role.yaml
Without this configuration, the operator is failing to re-conciliate the components and recreating them all the time.

Moreover, we should align the Status component by adding new fields such as "Applied" from KIECLOUD-376.
Finally, the names and replicas should be retrieved from the status.applied field instead.


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
```diff
@@ -42,8 +43,13 @@ public KieServerOperatorDeployment(Project project, NonNamespaceOperation<KieApp
...
         if (isReady()) {
             KieApp kieApp = kieAppClient.withName(OpenShiftConstants.getKieApplicationName()).get();
-            Server server = getAssociatedServerObject(kieApp);
+
+            Spec appliedSpec = kieApp.getStatus().getApplied();
+            Server server = getAssociatedServerObject(appliedSpec);
             server.setReplicas(instances);
+
+            // Update current spec
+            kieApp.setSpec(appliedSpec);
```
Sgitario, line 52: I did this way to make clear that we're getting an immutable copy of the spec and using it to update the current spec of the component.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
```diff
@@ -108,6 +118,25 @@ private void createRoleBindingsInProject(Project project) {
...
+    private void createClusterRoleInProject(Project project) {
+        logger.info("Creating cluster role in project '" + project.getName() + "' from " + OpenShiftResource.CLUSTER_ROLE.getResourceUrl().toString());
+        ClusterRole role = project.getOpenShiftAdmin().rbac().clusterRoles().load(OpenShiftResource.CLUSTER_ROLE.getResourceUrl()).get();
+        project.getOpenShiftAdmin().rbac().clusterRoles().inNamespace(project.getName()).createOrReplace(role);
+    }
+
+    private void createClusterRoleBindingsInProject(Project project) {
+        logger.info("Creating cluster role bindings in project '" + project.getName() + "' from " + OpenShiftResource.CLUSTER_ROLE_BINDING.getResourceUrl().toString());
+        ClusterRoleBinding roleBinding = project.getOpenShiftAdmin().rbac().clusterRoleBindings().load(OpenShiftResource.CLUSTER_ROLE_BINDING.getResourceUrl()).get();
+        roleBinding.getMetadata().setName(OPERATOR_DEPLOYMENT_NAME + project.getName());
```
Sgitario, line 130: This is to avoid collisions with other tests. 
The problem with this is that we need to remove the object when we delete the project (see the undeploy method above) as I confirmed that this is not done by Openshift when we delete the project.

Sgitario:
I ran some tests successfully.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
jakubschwan: here as we are only changing the replicas, I think it's ok to change only Spec and we not need to get all generated stuff that's it in applied. The default properties should be applied again.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/Status.java
```diff
@@ -27,6 +27,8 @@
     private String consoleHost;
     private Deployments deployments;
     private String phase;
+    private String version;
+    private Spec applied;
```
jakubschwan, line 31: just minor change request, can you move applied on the first position so the it's sorted. It's also better for checking of this Status class with status in KieApp

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
```diff
@@ -108,6 +118,25 @@ private void createRoleBindingsInProject(Project project) {
         project.getOpenShiftAdmin().rbac().roleBindings().inNamespace(project.getName()).create(roleBinding);
     }
 
+    private void createClusterRoleInProject(Project project) {
+        logger.info("Creating cluster role in project '" + project.getName() + "' from " + OpenShiftResource.CLUSTER_ROLE.getResourceUrl().toString());
+        ClusterRole role = project.getOpenShiftAdmin().rbac().clusterRoles().load(OpenShiftResource.CLUSTER_ROLE.getResourceUrl()).get();
+        project.getOpenShiftAdmin().rbac().clusterRoles().inNamespace(project.getName()).createOrReplace(role);
```
jakubschwan, line 124: Cluster role is shared and if this role already exists, then is thrown exception from Kubernetes client.
```suggestion
        try {
            project.getOpenShiftAdmin().rbac().clusterRoles().inNamespace(project.getName()).create(clusterRole);
        } catch (KubernetesClientException e) {
            if (e.getMessage().contains("already exists")) {
                logger.warn("Cluster role " + OpenShiftResource.CLUSTER_ROLE.getResourceUrl().toString() + " already exists.");
            } else {
                throw new RuntimeException("Exception when creating cluster role from " + OpenShiftResource.CLUSTER_ROLE.getResourceUrl().toString(), e);
            }
        }
```

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
```diff
@@ -108,6 +118,25 @@ private void createRoleBindingsInProject(Project project) {
...
+    }
+
+    private void createClusterRoleBindingsInProject(Project project) {
+        logger.info("Creating cluster role bindings in project '" + project.getName() + "' from " + OpenShiftResource.CLUSTER_ROLE_BINDING.getResourceUrl().toString());
+        ClusterRoleBinding roleBinding = project.getOpenShiftAdmin().rbac().clusterRoleBindings().load(OpenShiftResource.CLUSTER_ROLE_BINDING.getResourceUrl()).get();
+        roleBinding.getMetadata().setName(OPERATOR_DEPLOYMENT_NAME + project.getName());
+        roleBinding.getMetadata().setNamespace(project.getName());
+        roleBinding.getSubjects().forEach(subject -> subject.setNamespace(project.getName()));
```
jakubschwan, line 132: we can use projectName or getNamespace() instead of project.getName
```suggestion
        roleBinding.getMetadata().setName(OPERATOR_DEPLOYMENT_NAME + projectName);
        roleBinding.getMetadata().setNamespace(projectName);
        roleBinding.getSubjects().forEach(subject -> subject.setNamespace(projectName));
```

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
```diff
@@ -64,15 +70,15 @@ public void waitForScale() {
...
+    private Server getAssociatedServerObject(Spec spec) {
+        return Arrays.asList(spec.getObjects().getServers()).stream().filter(s -> s.getName().equals(getServiceName()))
                                                                                  .findAny()
-                                                                                 .orElseThrow(() -> new RuntimeException("Server with name " + getServiceName() + " not found. Available server names are: " + getAvailableServerNames(kieApp)));
+                                                                                 .orElseThrow(() -> new RuntimeException("Server with name " + getServiceName() + " not found. Available server names are: " + getAvailableServerNames(spec)));
     }
 
-    private String getAvailableServerNames(KieApp kieApp) {
-        String serverNames = Arrays.asList(kieApp.getSpec().getObjects().getServers()).stream().map(s -> s.getName())
+    private String getAvailableServerNames(Spec spec) {
```
jakubschwan, line 79: I think that here we can use kieApp.getStatus().getApplied() as we are getting available server names

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
```diff
@@ -64,15 +70,15 @@ public void waitForScale() {
         }
     }
 
-    private Server getAssociatedServerObject(KieApp kieApp) {
-        return Arrays.asList(kieApp.getSpec().getObjects().getServers()).stream().filter(s -> s.getName().equals(getServiceName()))
+    private Server getAssociatedServerObject(Spec spec) {
+        return Arrays.asList(spec.getObjects().getServers()).stream().filter(s -> s.getName().equals(getServiceName()))
```
jakubschwan, line 74: Here I think we can use kieApp.getStatus().getApplied() as we are trying to get created objects

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
Sgitario: The problem is that we need to update only the servers with a concrete name (from getServiceName()). 
In most of the scenarios, the name is populated automatically by the operator and driven by the environment (and it's in the status.applied). So, if the name is not supplied in the scenario builder, we would be fine to set the replicas as you suggested. But the problem is that in some other scenarios, we set the name in scenario builder (to have kie-server1 and kie-server2 for example) and we would only need to update one, not all.
Therefore, to cope with the both scenarios, I decided to use the specification from the status.applied, set the replicas and update the kieapp.spec. 

Another approach could be to detect if we have named servers, if so, search by name, if not, we scale all the servers. This works but can be trickier.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
Sgitario: As discussed, we are using the createOrReplace method instead of catching up exceptions.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/Status.java
Sgitario: Ok...

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/OpenShiftOperatorScenario.java
Sgitario: Ok. Will use getNamespace() method.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
Sgitario: Spec comes from kieApp.getStatus().getApplied() already. See line 48 and 61.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
Sgitario: Spec comes from kieApp.getStatus().getApplied() already. See line 48 and 61.

PR #639, opened 2020-06-29
[BAQE-1420] Adjust tests for Monitoring K8S filesystem

jakubschwan:


test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/jbpm/TimerIntegrationTest.java
```diff
@@ -46,7 +47,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@Category({OperatorNotSupported.class, JBPMOnly.class}) // Because TimerServiceDataStoreRefreshInterval is not supported yet, Operator is skipped as scenario use different startup strategy than template.
+@Category({OperatorNotSupported.class, JBPMOnly.class, MonitoringK8sFs.class}) // Because TimerServiceDataStoreRefreshInterval is not supported yet, Operator is skipped as scenario use different startup strategy than template.
```
Sgitario, line 50: Is this correct? This integration test will be executed in Templates only and the MonitoringK8sFs is only for Operator.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/KieServerWebSocketScalingIntegrationTest.java
```diff
@@ -50,7 +51,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@Category({Baseline.class, OperatorNotSupported.class}) // Operator doesn't contain any scenario with controller based strategy by default
+@Category({Baseline.class, OperatorNotSupported.class, MonitoringK8sFs.class}) // Operator doesn't contain any scenario with controller based strategy by default
```
Sgitario, line 54: Is this correct? This integration test will be executed in Templates only and the MonitoringK8sFs is only for Operator.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/KieServerWithSmartRouterHttpScalingIntegrationTest.java
```diff
@@ -47,7 +48,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@Category({OperatorNotSupported.class}) // Because DroolsServerFilterClasses not supported yet, Operator skipped as there is discrepancy between deployment methods and Kie server connection, should be unified for 7.4  
+@Category({OperatorNotSupported.class, MonitoringK8sFs.class}) // Because DroolsServerFilterClasses not supported yet, Operator skipped as there is discrepancy between deployment methods and Kie server connection, should be unified for 7.4
```
Sgitario, line 51: Is this correct? This integration test will be executed in Templates only and the MonitoringK8sFs is only for Operator.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/survival/KieServerWithSmartRouterAndControllerSurvivalIntegrationTest.java
```diff
@@ -54,7 +55,7 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-@Category({JBPMOnly.class, OperatorNotSupported.class})
+@Category({JBPMOnly.class, OperatorNotSupported.class, MonitoringK8sFs.class})
```
Sgitario, line 58: Is this correct? This integration test will be executed in Templates only and the MonitoringK8sFs is only for Operator.

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/KieServerWebSocketScalingIntegrationTest.java
jakubschwan: It may looks bad (and I must admit that I didn't realize that), but I would leave this configuration here. Right now this will be skipped as it is not supported in Operator, but in future we may allow some of the tests for Operator and we could forgot to set MonitoringK8sFs tests.
If we see that this configuration is in conflict in tests, we can remove it later.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/Build.java
```diff
@@ -86,4 +91,21 @@ public String getExtensionImageInstallDir() {
     public void setExtensionImageInstallDir(String extensionImageInstallDir) {
         this.extensionImageInstallDir = extensionImageInstallDir;
     }
+
+    public void addWebhooks(Webhooks webhooks) {
```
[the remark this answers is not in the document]
jakubschwan: Yes, they are needed to correctly set up KieApp used in tests. They were some changes for KieApp that caused test failures because not correctly generated KieApp

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/Build.java
```diff
@@ -86,4 +91,21 @@ public String getExtensionImageInstallDir() {
...
+        this.webhooks.add(webhooks);
+    }
+
+    public void addWebhooks(List<Webhooks> webhooks) {
+        this.webhooks.addAll(webhooks);
+    }
+
+    public Webhooks[] getWebhooks() {
```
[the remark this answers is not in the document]
jakubschwan: Followed same pattern that we are using in operator model

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/scaling/KieServerWebSocketScalingIntegrationTest.java
Sgitario: Ok ok

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/Build.java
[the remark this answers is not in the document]
Sgitario: Do you mean that this is used internally when marshalling/unmarshalling the Build instance?

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/Build.java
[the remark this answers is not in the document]
Sgitario: I found that we're setting collections in two different ways in the model:
- Using the getter/setter of arrays (the normal way). In [Status.java](https://github.com/kiegroup/kie-cloud-tests/blob/26e6e9f929a60df11b9b595f0b94d7c287e30574/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/Status.java#L45)
- Using addEnv, addEnvs, getEnv, getEnv (first time I saw this...). In several objects like [SmartRouter.java](https://github.com/kiegroup/kie-cloud-tests/blob/26e6e9f929a60df11b9b595f0b94d7c287e30574/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/SmartRouter.java#L31).

From my point of view, we should use only the getter/setter of the array/list, and maybe raise a ticket to do the same in all the collections. What do you think?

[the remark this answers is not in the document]
jakubschwan: for adding Env variables it make sense to leave it as it is right now. Env parameter can be added for different configuration and we do not want everytime create a new array (as we do not know the size of the array, and with adding to existing on we can get exception)
Right now for webhooks we probably do not need to add there list or one Webhook. I'll remove addWebhooks()

PR #647, opened 2020-07-06
BAQE-1467: Fix Upgrade Operator tests

Sgitario:
JIRA: https://issues.redhat.com/browse/BAQE-1467
Description:
Fix upgrade tests by:
- Checking whether the "applied" field exists to get the replicas number
- Checking the image reference via the provided sha256

### DIFF
--- a/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
+++ b/framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
@@ -16,6 +16,7 @@
 package org.kie.cloud.openshift.operator.deployment;
 
 import java.util.Arrays;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import io.fabric8.kubernetes.client.dsl.NonNamespaceOperation;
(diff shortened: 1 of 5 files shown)


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/deployment/KieServerOperatorDeployment.java
```diff
@@ -58,7 +59,10 @@ public void scale(int instances) {
...
         KieApp kieApp = kieAppClient.withName(OpenShiftConstants.getKieApplicationName()).get();
 
-        Integer replicas = getAssociatedServerObject(kieApp.getStatus().getApplied()).getReplicas();
+        Integer replicas = Optional.ofNullable(kieApp.getStatus())
+                                   .map(status -> status.getApplied())
+                                   .map(applied -> getAssociatedServerObject(applied).getReplicas())
+                                   .orElse(null);
         if (replicas == null) {
             // If replicas are not set in custom resource then get them from deployment config.
             replicas = getOpenShift().getDeploymentConfig(getServiceName()).getSpec().getReplicas();
```
jakubschwan, line 68: As we are using Optional here, can be code from if blocked moved to orElse part? So we would not need to return null to Integer replicas and then check it again for null.

Sgitario: Good point! Already changed in all the operator deployments.

PR #651, opened 2020-07-07
BAQE-1470: Fix External Database tests using Operator

Sgitario:
JIRA: https://issues.redhat.com/browse/BAQE-1470
Description: Use JDBC URL instead of HOST for external databases.


Sgitario:
Confirmed that the jenkins job is now working fine with these changes.

PR #668, opened 2020-09-02
[7.39.x] BAQE-1520: Fix DbSurvivalIntegrationTest tests in Operator

Sgitario:
https://issues.redhat.com/browse/BAQE-1520


Sgitario:
@jakubschwan can you take a look into this PR? I used this approach as I saw that this environment was in use in many other places, but the permanent storage was only needed in the DbSurvivalIntegrationTest.

jakubschwan (approved):

PR #669, opened 2020-09-02
[7.39.x] BAQE-1519: Fail fast when OC command is not installed

Sgitario:
https://issues.redhat.com/browse/BAQE-1519
Some tests return that some X service is not installed. The root cause is that the OC client was not installed. These changes are to make the tests fail fast when this happened.


Sgitario:
@jakubschwan can you take a look into this? If you agree with the solution, I will create another PR for the master.

jakubschwan (approved):

PR #681, opened 2020-10-05
Stabilizatoin of OCP smoke tests

jakubschwan:


test-cloud/test-cloud-common/src/main/java/org/kie/cloud/tests/common/client/util/WorkbenchUtils.java
```diff
@@ -33,28 +40,60 @@
...
+                                                                                         .interval(TimeUnit.SECONDS, 5)
+                                                                                         .waitFor();
+                    logger.info("Container after cloning timeout found");
+                } catch (NotFoundException e) {
+                    handleNotFoundException(workbenchClient, e);
+                } catch (WaiterException e) {
+                    handleWaiterException(tries, e);
+                }
+            }
+        }
```
Sgitario, line 96: I think this logic should be part of the cloneRepository method. 
Also, as the previous comment, I would prefer using the AwaitilityUtils for retries.

test-cloud/test-cloud-common/src/main/java/org/kie/cloud/tests/common/client/util/WorkbenchUtils.java
```diff
@@ -65,6 +104,49 @@ private static CloneProjectRequest createCloneProjectRequest(String repositoryUr
         return cloneProjectRequest;
     }
 
+    private static void handleNotFoundException(WorkbenchClient workbenchClient, Exception originalException) {
+        logger.debug("Exception handled when workbench client throws 404 when is tried to search for all spaces.");
+        boolean workbenchWorks = false;
+        for (int i = 0; i < 10; i++) {
```
Sgitario, line 110: As the previous comment, I would prefer using the AwaitilityUtils for retries.

test-cloud/test-cloud-remote/src/main/java/org/kie/cloud/integrationtests/runners/SmokeTestRunner.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    protected final void runTestUnit(Statement statement, Description description, RunNotifier notifier) {
+        EachTestNotifier eachNotifier = new EachTestNotifier(notifier, description);
+        eachNotifier.fireTestStarted();
+        try {
+            statement.evaluate();
+        } catch (AssumptionViolatedException e) {
+            eachNotifier.addFailedAssumption(e);
+        } catch (Throwable e) {
```
Sgitario, line 74: I should not retry on any kind of exception, but only when one exception was thrown. We could create a custom exception for this.
Also, if we finally use the AwaitilityUtils, we could use the .conditionEvaluationListener method to raise this custom exception only when required. 
What do you think?

test-cloud/test-cloud-remote/src/test/java/org/kie/cloud/integrationtests/smoke/WorkbenchKieServerScenarioIntegrationTest.java
```diff
@@ -41,6 +43,7 @@
 import org.kie.server.controller.client.KieServerControllerClient;
 
 @Category(Smoke.class)
+@RunWith(SmokeTestRunner.class)
```
Sgitario, line 46: Adding the category and the smoke test runner seems to me like a bit verbose. 
Maybe we could use Maven to configure the tests annotated with smoke category by using the smoke test runner.
Or perhaps we could use this custom test runner in all the tests (if and only if we only retry when a deployment timeout exception was raised). What do you think?

Sgitario (commented):
Thanks! Jakub, mainly two comments:
- From my point of view, it would be clearer if we use the AwailabilityUtils instead of looping for a retries integer. (See comments)
- About the SmokeTestRunner, I think we should only retry when a new custom exception is thrown (a Deployment Timeout exception or alike). Also, what about using this test runner always?

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/resource/impl/ProjectImpl.java
```diff
@@ -206,11 +207,20 @@ private OpenShiftBinary openShiftBinaryClient() {
...
+                try {
+                    oc = Optional.of(OpenShifts.masterBinary(this.getName()));
+                    oc.get().login(OpenShiftConstants.getOpenShiftUrl(), OpenShiftConstants.getOpenShiftAdminUserName(),
+                                   OpenShiftConstants.getOpenShiftAdminPassword());
+                    break;
+                } catch (HttpsException ex) {
+                    logger.warn("Was caught exception from OpenShifts on " + attempt
+                            + " attempt. Trying to get master binaries again.", ex);
+                }
+            }
```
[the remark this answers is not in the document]
sutaakar: Another suggestion - what about dropping Optional? I don't see any relevant use case for it here.
It would also simplify method execution flow.

test-cloud/test-cloud-common/src/main/java/org/kie/cloud/tests/common/client/util/WorkbenchUtils.java
```diff
@@ -65,6 +104,49 @@ private static CloneProjectRequest createCloneProjectRequest(String repositoryUr
...
+                    logger.debug("Container URL works, but space was not found. Waiting for 30 seconds and trying again.");
+                    waitFor30Seconds();
+                }
+            } catch (NotFoundException ee) {
+                logger.debug("Container URL do not work. Waiting for 30 seconds and trying again.", ee);
+                waitFor30Seconds();
+            }
+        }
+        if (workbenchWorks) {
+            logger.info("After few retries Workbench client works. NotFoundException can be ignored.");
```
sutaakar, line 126:
```suggestion
            logger.info("After " + (i+1) + " retries Workbench client works. NotFoundException can be ignored.");
```

test-cloud/test-cloud-common/src/main/java/org/kie/cloud/tests/common/client/util/WorkbenchUtils.java
```diff
@@ -65,6 +104,49 @@ private static CloneProjectRequest createCloneProjectRequest(String repositoryUr
...
+                                       originalException);
+        }
+    }
+
+    private static void waitFor30Seconds() {
+        try {
+            Thread.sleep(TimeUnit.SECONDS.toMillis(30));
+        } catch (InterruptedException e1) {
+            // TODO Auto-generated catch block
+            e1.printStackTrace();
```
sutaakar, line 146:
```suggestion
            Thread.currentThread().interrupt();
            throw new RuntimeException("Interrupted while waiting.", e1);
```

framework-cloud/framework-openshift/src/main/java/org/kie/cloud/openshift/resource/impl/ProjectImpl.java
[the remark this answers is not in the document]
Sgitario: We're building a test framework. Why don't use it and avoid to reinvent the wheel here? About not catching all the exceptions, we can configure it to do it (see options in https://github.com/awaitility/awaitility/wiki/Usage)

test-cloud/test-cloud-common/src/main/java/org/kie/cloud/tests/common/client/util/WorkbenchUtils.java
Sgitario: Don't you think that ensuring the repository is created in any kind of environments and methods is a good thing? In the best case (non HA environments) we will only lose few time. Also, we could spot issues earlier (in case of for some reason, the repository is not created). This is just my opinion.

test-cloud/test-cloud-common/src/main/java/org/kie/cloud/tests/common/client/util/WorkbenchUtils.java
Sgitario: Why did we need to either print or interrupt the thread? :/

sutaakar: @jakubschwan The InterruptedException is thrown for example in case the process is getting terminated. In such case we should stop immediately and don't continue with execution flow.
@Sgitario I read somewhere that this is a standard handling of InterruptedException. So I prefer to follow it.

sutaakar (approved):

PR #694, opened 2020-12-01
Fix workbench memory limit for BC HA

tomasdavidorg:


tomasdavidorg:
@jakubschwan Could you please take a look?

PR #764, opened 2022-01-12
[RHPAM-4115] Add custom hostname integration test

jakubschwan:


framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/CommonConfig.java
```diff
@@ -64,6 +66,10 @@ public String getDbPassword() {
         return dbPassword;
     }
 
+    public Boolean getDisableSsl() {
```
radtriste, line 69: shouldn't it be `isDisableSsl` ?

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/AbstractOpenshiftScenarioBuilderOperator.java
```diff
@@ -28,4 +29,69 @@ protected ExternalDeploymentFactory getExternalDeploymentFactory() {
...
+
+    protected void checkHttpKieServerRouteConfig(boolean kieServerHostnameSet, KieApp kieApp) {
+        if (kieServerHostnameSet) {
+            throw new IllegalStateException("Kie Server hostname has been already set. Update your scenario.");
+        }
+        if (!isDisableSslNull(kieApp) && isDisableSslFalse(kieApp)) {
+            throw new IllegalStateException("SSL is enabled. It is not possible to configure http hostname.");
+        }
+
+        if (isDisableSslNull(kieApp)) {
```
radtriste, line 52: not sure you need such a check.
you check already: `if not null and false`
then you can just do `kieApp.getSpec().getCommonConfig().setDisableSsl(true);` wihtout the `if`

radtriste (approved):
minor comments

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/model/components/CommonConfig.java
jakubschwan: No, it's part of KieApp model and here are required getter and setters for configuration.

framework-cloud/framework-openshift-operator/src/main/java/org/kie/cloud/openshift/operator/scenario/builder/AbstractOpenshiftScenarioBuilderOperator.java
jakubschwan: yes this check is needed in case disables ssl parameter is not set. This parameter is used to set up the url protocol.

